PR #11, opened 2023-02-02
Fix misc bugs and issues

FabienRoger:


Butanium:
Won't that break import using `elk.` ? Like 
```py
# generation_main.py
from elk.utils_generation.generation import create_records, create_hiddenstates
```

FabienRoger:
> Won't that break imports using `elk.` ? Like
> 
> ```python
> # generation_main.py
> from elk.utils_generation.generation import create_records, create_hiddenstates
> ```

To `cd` in elk? No, because elk is pip installed module. Using `elk.` works anywhere.

elk/train.py
```diff
@@ -54,6 +57,7 @@ def train(args):
 
     # save models
     # TODO: use better filename for the pkls, so they don't get overwritten
+    Path(args.trained_models_path).mkdir(parents=True, exist_ok=True)
```
norabelrose, line 60: nit: `args.trained_models_path` is already a `Path` so we don't have to make it one here

norabelrose (approved):
LGTM, I have one little code style nit about line 60 of `train.py` but that could be changed in a later PR or something

PR #12, opened 2023-02-02
Add Lbfgs

FabienRoger:
Fix https://github.com/EleutherAI/elk/issues/1


elk/utils_evaluation/ccs.py
```diff
@@ -217,3 +203,74 @@ def fit(self, data: list, label):
...
+            # get the corresponding loss
+            loss = self.get_loss(p0, p1)
+
+            # update the parameters
+            optimizer.zero_grad()
+            loss.backward()
+            optimizer.step()
+
+            # with torch.no_grad():
+            #     theta /= torch.norm(theta)
```
norabelrose, line 233: This should probably just be deleted if we're not using it?

elk/utils_evaluation/ccs.py
```diff
@@ -217,3 +203,74 @@ def fit(self, data: list, label):
...
+            # get the corresponding loss
+            loss = self.get_loss(p0, p1)
+
+            loss += l2 * torch.norm(theta) ** 2 / 2
+
+            # update the parameters
+            loss.backward()
+
+            # with torch.no_grad():
+            #     theta /= torch.norm(theta)
```
norabelrose, line 271: Same as above

norabelrose (commented):
Looks good overall, the main thing I'd like to see changed is changing `--use_lbfgs` to `--optimizer` with two string options. But maybe Walter has a different opinion.

elk/train.py
```diff
@@ -37,7 +40,9 @@ def train(args):
     print("done training classification model")
 
     print("train ccs model")
-    ccs_model = CCS(verbose=True)
+    ccs_model = CCS(
+        verbose=True, weight_decay=args.weight_decay, use_lbfgs=args.use_lbfgs
```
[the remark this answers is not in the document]
FabienRoger: Agreed, I'm adding that

requirements.txt
```diff
@@ -1,5 +1,5 @@
 transformers==4.25.1
-promptsource==0.2.3
+git+https://github.com/bigscience-workshop/promptsource.git
```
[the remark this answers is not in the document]
FabienRoger: I made it coherent in the toml file

elk/utils_evaluation/parser.py
```diff
@@ -44,14 +44,25 @@ def get_args(default_config_path=Path(__file__).parent / "default_config.json"):
...
+    )  # TODO: explain this. Never used.
     parser.add_argument("--seed", type=int, default=0)
     parser.add_argument(
         "--model_device",
         type=str,
         default="cuda",
         help="What device to load the model onto: CPU or GPU or MPS.",
     )
+    parser.add_argument(
+        "--use_lbfgs", action="store_true", help="Use lbfgs optimizer to train CCS."
```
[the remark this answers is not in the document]
FabienRoger: I used an enum for type safety. Literal["adam", "lbfgs"] can't be used in argparse.
But maybe you think it's better to use str and then assert that the optimizer is correct?

elk/utils_evaluation/ccs.py
FabienRoger: Done

elk/utils_evaluation/parser.py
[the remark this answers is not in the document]
norabelrose: You can use `choices=("adam", "lbfgs")` in `argparse`. I tend not to use enums because they're verbose— you have to type `Optimizer.adam` instead of just `"adam"`. For now I'd recommend just using strings + an `assert` statement. But maybe @lauritowal disagrees.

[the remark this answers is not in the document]
FabienRoger: Done. Also, the assert is done by argparse in the parser, and I had already an assert in the training code for ccs, so I guess no additional assert is needed.

elk/utils_evaluation/parser.py
```diff
@@ -44,14 +44,25 @@ def get_args(default_config_path=Path(__file__).parent / "default_config.json"):
     )
     parser.add_argument("--language_model_type", type=str, default="encoder")
     parser.add_argument("--layer", type=int, default=-1)
-    parser.add_argument("--zero", type=str, default="results")
+    parser.add_argument(
+        "--zero", type=str, default="results"
+    )  # TODO: explain this. Never used.
```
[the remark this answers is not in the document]
lauritowal: yeah, it's from the old code... The two parser files have to be cleaned up more

elk/utils_evaluation/ccs.py
```diff
@@ -217,3 +213,68 @@ def fit(self, data: list, label):
     def score(self, data: list, label, getloss=False):
         self.validate_data(data)
         return self.get_accuracy(self.best_theta, data, label, getloss)
+
+    def train_loop_full_batch(self, x0, x1, theta):
```
lauritowal, line 210: It's not really a batch, no? We are running through self.num_epochs. Maybe, let's just call it train_loop_adam for now?

lauritowal (changes requested):
It looks mostly good to me.  Just as a small note: Next time, it would be maybe cleaner to create two pull-requests. One for the cleaning up and one for the actual feature / fix you want to add

elk/utils_evaluation/ccs.py
FabienRoger: Noted!

elk/utils_evaluation/ccs.py
```diff
@@ -1,6 +1,8 @@
+from typing import Literal
 import numpy as np
 import torch
 import matplotlib.pyplot as plt
+from enum import Enum
```
lauritowal, line 5: Is it used somewhere?

norabelrose: I don't think so, it's a vestige from earlier when Fabien was using an `Enum`

norabelrose (approved):
LGTM

lauritowal (approved):

Issue #7, opened 2023-02-01
Enable more general probes


FabienRoger:
- MLP
- Linear probes with constraints

PR #13, opened 2023-02-02
Make CCS use any kind of probes

FabienRoger:
Fix https://github.com/EleutherAI/elk/issues/7
Fix https://github.com/EleutherAI/elk/issues/18


norabelrose:
I'll review this after #12 gets merged

elk/utils_evaluation/ccs.py
```diff
@@ -11,47 +13,27 @@ def __init__(
...
-        Initializes the parameters of the model.
+        Initializes the probe of the model.
 
         Returns:
-            numpy.ndarray: The initialized parameters of the model.
+            Probe: The initialized probe of the model.
         """
-        init_theta = np.random.randn(self.d).reshape(1, -1)
-        init_theta = init_theta / np.linalg.norm(init_theta)
-        return init_theta
```
lauritowal, line 32: I think we should keep the initialization of theta for now (at least for Adam). Just for reproducing the actual results of the paper. It could be of course that it doesn't make any difference, and we remove it later.

elk/utils_evaluation/ccs.py
```diff
@@ -80,25 +62,24 @@ def get_loss(self, p0, p1):
 
         return consistency_loss + confidence_loss
 
-    # return the probability tuple (p0, p1)
-
-    def transform(self, data: list, theta_np=None):
-        if theta_np is None:
-            theta_np = self.best_theta
-        z0 = torch.tensor(self.add_ones_dimension(data[0]).dot(theta_np.T))
-        z1 = torch.tensor(self.add_ones_dimension(data[1]).dot(theta_np.T))
```
lauritowal, line 89: Similarly, maybe we should keep add_ones_dimension for now. Just for reproducing the actual results of the paper. It could be of course that it doesn't make any difference, and we remove it later in a different pull request.

pyproject.toml
```diff
@@ -14,7 +14,7 @@ dependencies = [
     "numpy==1.23.5",
     "pandas==1.5.1",
     "transformers==4.25.1",
-    "promptsource==0.2.3",
+    "git+https://github.com/bigscience-workshop/promptsource.git",
```
lauritowal, line 17: Any reason, of why we want always the latest version of promptsource? Making sure that we have all the datasets? (Just asking, because this could reguallry maybe break things?)
(I have also a [_custom_ promptsource](https://github.com/NotodAI-Research/promptsource) to use "unofficial" datasets, which we might want to include into our project as submodule maybe.)

lauritowal (changes requested):

pyproject.toml
FabienRoger: 0.2.3 only work with python 3.7

FabienRoger: but maybe your promptsource doens't? I can switch to yours if it's better

norabelrose: I think we definitely shouldn't list `promptsource==0.2.3` because it literally doesn't work on the Python versions that we recommend/require (3.9 and above). I'm fine with either `git+https://github.com/bigscience-workshop/promptsource.git` or `git+https://github.com/NotodAI-Research/promptsource.git`.

lauritowal: Let's link to git+https://github.com/bigscience-workshop/promptsource.git for now. I'll clean up my promptsource a bit and linkt to that later

elk/utils_evaluation/ccs.py
FabienRoger: Done (in the probe with the original init).

elk/utils_evaluation/ccs.py
FabienRoger: Agreed, I've added a probe which is initialized just like the original CCS. Thanks for catching that.

FabienRoger:
Note: this change makes loading old pickled CCS fail. Maybe we should go away from pickle soon (in this PR?).

elk/utils_evaluation/ccs.py
```diff
@@ -1,57 +1,46 @@
...
+
+class CCS(torch.nn.Module):
     def __init__(
         self,
+        d,
         verbose=False,
         include_bias=True,
-        num_epochs=1000,
-        num_tries=10,
-        learning_rate=1e-2,
```
lauritowal, line 13: Shouldn't we leave that stuff in the constructor, if want it to be like scikitlearn:
See for example:
https://scikit-learn.org/stable/modules/generated/sklearn.neural_network.MLPClassifier.html

learning_rate, batch_size etc. are set in the constructor there

norabelrose:
@FabienRoger I'm in favor of just turning this PR into the one that fixes #18, in which case we'd use `torch.save` on the module state dict

elk/utils_evaluation/ccs.py
```diff
@@ -170,29 +129,48 @@ def visualize(self, losses, accs):
...
     # switch back to 1000 + 10
-    def fit(self, data: list, label):
+    def fit(
+        self,
+        data: list,
+        label,
+        num_epochs=1000,
+        num_tries=10,
+        learning_rate=1e-2,
+        weight_decay=0.01,
```
lauritowal, line 139: see comment above

lauritowal (changes requested):

elk/utils_evaluation/ccs.py
FabienRoger: My bad, I thought sklearn API was the same as Keras' (which asks for learning params in the fit function). I think we should revert the very last commit.

norabelrose: I disagree, actually. We don't have to be _exactly_ like `sklearn` and this aspect of their API doesn't really make any sense, especially for a PyTorch module. It just forces you to create a bunch of instance attributes that don't need to be there otherwise. And what if you want to fit the model multiple times with different hyperparameters?

PR #19, opened 2023-02-02
implement test cases for reproducing the results

mariopenglee:
only dataframe comparison for now, need to add evaluations

### DIFF
--- a/elk/extraction/prompt_collator.py
+++ b/elk/extraction/prompt_collator.py
@@ -70,6 +70,9 @@ class PromptCollator(Dataset):
 
         self.dataset = self.dataset.shuffle(seed=seed)
         if max_examples:
+            num_examples = len(self.dataset)
+            if max_examples > num_examples:
+                max_examples = num_examples
             self.dataset = self.dataset.select(range(max_examples))
 
         self.label_column = label_column


elk/tests/test_extraction_output.py
```diff
@@ -1,17 +1,47 @@
...
+    )
+    current_performance = pd.read_csv(
+        "../evaluation_results/{}_confusion_0.csv".format(model_name)
+    )
+
+    # Compare the contents of the dataframes,
+    # make sure current performance are equal or better than the expected results
+    for index, row in benchmark_performance.iterrows():
+        assert row["accuracy"] <= current_performance.iloc[index]["accuracy"]
+        assert row["std"] <= current_performance.iloc[index]["std"]
```
lauritowal, line 42: I guess we should maybe add some +/- epsilon (e.g having value 3)  to increase the tolerance a bit (?)

norabelrose: Yeah probably. We should make sure the random seed is fixed and then just see empirically how big the discrepancy is with the current code

elk/tests/test_extraction_output.py
```diff
@@ -1,17 +1,66 @@
...
+    for ccs and logistic regression
+
+    generate result different method
+    """
+
+    # Train and evaluate the model
+    train_and_evaluate(model_name, dataset)
+
+    # Read the contents of the csv files into pandas dataframes
+    benchmark_performance = pd.read_csv(f"../test_data/{model_name}_confusion_0.csv")
```
lauritowal, line 52: To be able to reproduce the results of the paper, we want mostly  the prefix "normal" not "confusion"  --> See parser file. However, it might make sense to have an option to be able to select between the different prefixes for the test

elk/tests/test_extraction_output.py
```diff
@@ -1,17 +1,66 @@
...
+    current_performance = pd.read_csv(
+        f"../evaluation_results/{model_name}_confusion_0.csv"
+    )
+
+    # Compare the contents of the dataframes,
+    # make sure current performance is within threshold (epsilon) or better
+    epsilon = 3
```
lauritowal, line 59: it would be better if we pass epsilon, instead of hard-coding it here

elk/tests/test_extraction_output.py
```diff
@@ -1,17 +1,66 @@
...
+prefix = default_config["prefix"]
+models_layer_num = default_config["models_layer_num"]
+
+
+def train_and_evaluate(model_name, dataset):
+    """
+    Train and evaluate the model with the given dataset.
+    """
+    training_command = "python -m elk.train "
+    f"--model {model_name} --prefix normal --dataset {dataset} --num_data 1000 --seed 0"
```
lauritowal: Maybe it is better, if we call the functions from train.py and evaluate.py directly?

elk/tests/test_extraction_output.py
lauritowal, line 52: You might want to create this csv file (probably manually) and add it to the repo. It should contain all the different combinations from the table in the paper (for logistic regression and CCS only): https://ar5iv.labs.arxiv.org/html/2212.03827/assets/x8.png

lauritowal (changes requested):

PR #23, opened 2023-02-02
Task set random seed

lauritowal:
Simple solution for the random seed for now...

### DIFF
--- a/elk/utils_evaluation/utils_evaluation.py
+++ b/elk/utils_evaluation/utils_evaluation.py
@@ -4,6 +4,7 @@ import numpy as np
 import pandas as pd
 import json
 import time
+import torch
 
 from pathlib import Path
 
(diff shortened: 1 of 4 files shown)


elk/utils_evaluation/utils_evaluation.py
```diff
@@ -18,6 +19,14 @@
 models_layer_num = default_config["models_layer_num"]
 
 
+def set_random_seed(seed):
+    """
+    Set the random seed
+    """
+    random.seed(seed)
+    np.random.seed(seed)
```
norabelrose, line 27: Can you add `torch.manual_seed(seed)` here? We now use PyTorch randomness to initialize the probe weights

norabelrose (changes requested):
Just add `torch.manual_seed` and it should be good to merge

norabelrose:
Already solved by earlier commits

Issue #14, opened 2023-02-02
Story_cloze fails


FabienRoger:
This dataset is not recognized by huggingface

```
load raw dataset story-cloze from module.
Downloading and preparing dataset story_cloze/2016 to /home/ubuntu/.cache/huggingface/datasets/story_cloze/2016-data_dir=.%2Fdatasets%2Frawdata/0.0.0/45cead0538c3deb72d731a7990e60835c2c9c5d5d5b1e95a7dd47ccf593671e4...


Generating validation split:   0%|          | 0/1871 [00:00<?, ? examples/s][A
Iterating over prefixes::   0%|          | 0/1 [03:59<?, ?it/s]
Traceback (most recent call last):
  File "/home/ubuntu/miniconda3/envs/elk/lib/python3.9/site-packages/datasets/builder.py", line 1570, in _prepare_split_single
    for key, record in generator:
  File "/home/ubuntu/.cache/huggingface/modules/datasets_modules/datasets/story_cloze/45cead0538c3deb72d731a7990e60835c2c9c5d5d5b1e95a7dd47ccf593671e4/story_cloze.py", line 112, in _generate_examples
    with open(filepath, encoding="utf-8") as csv_file:
  File "/home/ubuntu/miniconda3/envs/elk/lib/python3.9/site-packages/datasets/streaming.py", line 69, in wrapper
    return function(*args, use_auth_token=use_auth_token, **kwargs)
  File "/home/ubuntu/miniconda3/envs/elk/lib/python3.9/site-packages/datasets/download/streaming_download_manager.py", line 445, in xopen
    return open(main_hop, mode, *args, **kwargs)
FileNotFoundError: [Errno 2] No such file or directory: '/home/ubuntu/elk/elk/datasets/rawdata/cloze_test_val__spring2016 - cloze_test_ALL_val.csv'
```

FabienRoger:
Probably due to these lines. Does someone know why they are here?
```python
if set_name != "story-cloze":
        raw_set = load_dataset(*getLoadName(set_name))
    else:
        raw_set = load_dataset(*getLoadName(set_name), data_dir="./datasets/rawdata")
```

haileyschoelkopf:
Story Cloze requires manual download after requesting the data from this form:
found link via: https://huggingface.co/datasets/story_cloze/blob/27f6d5766d833abe979f7b49457e9458c89b94fb/story_cloze.py#L52

form: 
[...]

PR #29, opened 2023-02-03
Add an informative error message for story-cloze

FabienRoger:
Fix https://github.com/EleutherAI/elk/issues/14


norabelrose (approved):
This is better than what we have now, although next week we should look at getting rid of the dedicated `data_dir` entirely

PR #30, opened 2023-02-03
Use train data for computing normalization

FabienRoger:


elk/train.py
```diff
@@ -30,14 +31,16 @@ def train(args):
         mode=args.mode,
         num_data=args.num_data,
     )
+    permutation = get_permutation(hidden_states)
+    normalized_hidden_states = normalize(hidden_states, permutation)
```
lauritowal, line 35: I guess you are right; it is cleaner when the normalization happens outside of the get_hidden_states function. Thanks

elk/utils_evaluation/utils_evaluation.py
```diff
@@ -58,14 +61,47 @@ def organize(hidden_states, mode):
...
+        else:
+            scale_factors = [
+                np.std(data[permutation[TRAIN_SPLIT_IDX]], axis=0)
+                for data, label in hidden_states
+            ]
+        hidden_states = [
+            (data * scale_factor, label)
+            for (data, label), scale_factor in zip(hidden_states, scale_factors)
+        ]
+    elif scale == "projection":
```
lauritowal, line 96: Maybe we should find a better name for "projection", aren't all the other normalization here not also projections?

lauritowal (changes requested):
Seems like deberta performs worse after these changes:

for boolq for it was 70% before, now:

```
model,prefix,method,prompt_level,train,test,accuracy,std,language_model_type,layer,loss
deberta-v2-xxlarge-mnli,normal,ccs,all,boolq,boolq,0.551499992609024,0.051188381401737586,encoder,-1.0,0.37969053238630296
deberta-v2-xxlarge-mnli,normal,lr,all,boolq,boolq,0.6075,0.06677387213573883,encoder,-1.0,0.0"
```

for imdb for it was 89% before, now:
```
model,prefix,method,prompt_level,train,test,accuracy,std,language_model_type,layer,loss
deberta-v2-xxlarge-mnli,normal,ccs,all,imdb,imdb,0.8411666552225748,0.15101562370137744,encoder,-1.0,0.8821464618047078
deberta-v2-xxlarge-mnli,normal,lr,all,imdb,imdb,0.9106666666666665,0.07123591478710409,encoder,-1.0,0.0
```

We should check what is happening there (?)

elk/utils_evaluation/utils_evaluation.py
```diff
@@ -58,14 +61,47 @@ def organize(hidden_states, mode):
...
-    if not scale:
-        return data
-    norm = np.linalg.norm(data, axis=1)
-    avgnorm = np.mean(norm)
-    return data / avgnorm * np.sqrt(data.shape[1])
+def normalize(
+    hidden_states,
+    permutation,
+    scale: Literal["original", "none", "elementwise", "projection"] = "original",
+    demean=True,
```
lauritowal, line 68: Add a boolean include_test_set flag to use the original normalization without splitting test and train set

lauritowal (changes requested):

elk/utils_evaluation/utils_evaluation.py
FabienRoger: I've renamed it to layernorm since this is the layernorm projection. Do you prefer this name?

FabienRoger:
> for boolq for it was 70% before, now:
I'm not finding that. I got 59% for both.
> for imdb for it was 89% before, now:
I'm not finding that. I got 0.88 for both runs.

lauritowal:
Are you sure that you are not loading an older model? What model type are you using? I'll try it out again, later.

FabienRoger:
deberta-v2-xxlarge-mnli,normal,ccs,all,boolq,boolq,0.5864999830722809,0.0585042759828347,encoder,-1.0,0.9317224085330963

FabienRoger:
I fixed a dumb bug on now the results are
On boolq, drop from 65% to 62%
On imdb, stays at 88%

norabelrose (approved):
LGTM

Issue #25, opened 2023-02-03
Stop using a separate `cache_dir` to save pretrained models


norabelrose:
I don't see a reason why we can't use the HuggingFace default `cache_dir`. The current behavior simply causes models to get downloaded and saved to disk needlessly when the user already has a copy in the primary `cache_dir`. Also the `models` folder is annoying and clutters your working directory.

PR #31, opened 2023-02-03
Support all HF models

neverix:
This PR removes checks for the model list and changes the config format.


elk/utils_evaluation/parser.py
```diff
@@ -67,13 +66,20 @@ def get_args(default_config_path=Path(__file__).parent / "default_config.json"):
     )
     args = parser.parse_args()
 
+    # Dereference shortcut
+    args.model = model_shortcuts.get(args.model, args.model)
```
lauritowal, line 70: Why the double args.model?

elk/utils_evaluation/utils_evaluation.py
```diff
@@ -32,7 +31,7 @@ def get_filtered_filenames(
     :param confusion: the confusion
     :param place: the place where the data is from
     """
-    files = map(lambda path: path.name, directory.iterdir())
+    files = map(lambda path: str(path.relative_to(directory)), directory.glob("**/"))
```
lauritowal, line 34: What was the issue with the previous line of code?

elk/default_config.json
```diff
@@ -11,14 +11,13 @@
...
   ],
-  "models": [
-    "t5-11b",
-    "unifiedqa-t5-11b",
-    "T0pp",
-    "gpt-j-6B",
-    "roberta-large-mnli",
-    "deberta-v2-xxlarge-mnli"
-  ],
+  "model_shortcuts" : {
```
lauritowal, line 14: As it is right now, it will lead to have slashes in the directory name. Not sure if this is desirable. Maybe we should avoid that, or replace the string when creating the folders.

lauritowal (changes requested):

elk/utils_evaluation/parser.py
norabelrose: `.get()` takes an optional second parameter which is the default value returned when the key doesn't exist in the dictionary. So if the model string isn't in the shortcuts dictionary, it's interpreted as a raw HF model string

elk/utils_evaluation/utils_evaluation.py
norabelrose: The new code searches for files _recursively_ while the previous code only looked at the top level. I think.

norabelrose:
I fixed some unbound variable errors. Also I couldn't help myself and went ahead and fixed #25 because it was annoying me.

elk/default_config.json
norabelrose: Hmmm it may actually be fine if we're consistent about it? Like you just end up creating an `EleutherAI` parent folder. But idk.

elk/utils_evaluation/utils_evaluation.py
neverix: That's it, a path like `microsoft/deberta` becomes a two-level directory. We could replace slashes with some other character but imo that wouldn't be worth keeping it up everywhere.

lauritowal:
For Deberta I get the following error:
```
Traceback (most recent call last):
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/wombat_share/laurito/elk/elk/generation_main.py", line 55, in <module>
    create_hiddenstates(model, tokenizer, name_to_dataframe, args)
  File "/home/wombat_share/laurito/elk/elk/utils_generation/generation.py", line 108, in create_hiddenstates
    hidden_states = calculate_hidden_state(
  File "/home/wombat_share/laurito/elk/elk/utils_generation/generation.py", line 28, in calculate_hidden_state
    assert (
AssertionError: BERT does not have decoder. Relevant args: model=microsoft/deberta-v2-xxlarge-mnli, states_location=decoder.
```

Issue #35, opened 2023-02-04
Use better filenames for the pkls so they don't get overwritten


QuintinPope:

FabienRoger:
I suggest passing a "run name" as argument to avoid having super long file names specifying every single parameter of the runs. The run name would prefix both the pkl file names & the results csv files.

Is there a better third option? Or do people usually eat the super long file names to avoid having to track how they called their runs? (The full parameters and metadata of each run would be saved in a json file).

PR #42, opened 2023-02-07
Add custom run name for train & eval

FabienRoger:
Add the --run-name flag. Fix https://github.com/EleutherAI/elk/issues/35

Also add a tiny bug fix


norabelrose (approved):
LGTM

Issue #26, opened 2023-02-03
Rename "generation" to "extraction" of hidden states


norabelrose:
All throughout the code, the process of running inference on the model and collecting the hidden states is called "generation." I find this very confusing and counterintuitive because this term is used in HF transformers to refer to autoregressive _text_ generation. But we're not generating text. I think the verb "extract" is much less confusing and should be used in this codebase instead.

Issue #9, opened 2023-02-01
Turn the extraction and evaluate scripts into CLI commands


norabelrose:
I want to be able to do `elk extract <args>` and `elk evaluate <args>` after pip installing the package

PR #44, opened 2023-02-08
Mother of all refactors

### LINKED ISSUE #32: Move generate.sh, evaluate.sh, created directories outside of package

norabelrose:
This PR is a full rewrite of the code used to extract hidden states from models, dramatically simplifying and generalizing it.

High level overview of some of the big changes:
- Arbitrary HuggingFace models and datasets are supported, fixing #31
- New CLI, with commands like `elk extract` and `elk train` available upon installation, fixing #9 
- Hidden state "generation" is renamed to "extraction" throughout the code, fixing #26
- Instead of saving generated prompts with Pandas and saving them in a folder inside the root repo directory, we now generate them on-the-fly with a new `PromptCollator` class.
- Hidden states and probes are now saved in subfolders inside the `~/.cache/elk/` directory. The subfolders are named with the MD5 hash of the CLI args unless a custom name is provided (fixes #32)
- We now use use [a forked version](https://github.com/NotodAI-Research/promptsource) of the `promptsource` library which includes some of the extra prompts used by the original CCS codebase. We also changed the `setup.py` in the forked repo so that it doesn't complain when you try to install on Python 3.10.
- Hidden states are saved for _all_ layers of the transformer by default, and probes are trained and evaluated for each layer of the network by default
- We now use the train/test split built into the HF dataset itself when available, instead of taking the train set and trying to split it ourselves (which can introduce subtle leakages)

There are quite likely bugs in the current version, and there's still a couple things that we need to add back to this code that were in the old code (like confusion prompt prefixes), but it's gotten to the point where I'd like people besides me and @lauritowal to look at the code and provide feedback.


elk/train.py
```diff
@@ -1,79 +1,148 @@
...
+        lr_models.append(lr_model)
+        ccs_models.append(ccs_model)
+
+    ccs_models.reverse()
+    lr_models.reverse()
+    return lr_models, ccs_models
 
 
 if __name__ == "__main__":
-    args = get_args(default_config_path=Path(__file__).parent / "default_config.json")
```
FabienRoger, line 68: This parser is wrong? It does not take model or dataset. Running the train.sh script fails.

FabienRoger (commented):
lgtm, except the parse arguments of the train script

elk/train.py
```diff
@@ -1,79 +1,148 @@
...
+            "Weight decay for CCS when using Adam. Used as L2 regularization in LBFGS."
+        ),
+    )
+    args = parser.parse_args()
+
+    # Default to CUDA iff available
+    if args.device is None:
+        import torch
+
+        args.device = "cuda" if torch.cuda.is_available() else "cpu"
```
FabienRoger, line 70: I mean this is wrong

FabienRoger, line 70: I mean this is wrong

elk/train.py
```diff
@@ -1,79 +1,148 @@
...
+    )
+    parser.add_argument("--seed", type=int, default=0)
+    parser.add_argument(
+        "--weight-decay",
+        type=float,
+        default=0.01,
+        help=(
+            "Weight decay for CCS when using Adam. Used as L2 regularization in LBFGS."
+        ),
+    )
```
FabienRoger, line 133: This is wrong, right? No model name, no dataset, ...

elk/train.py
```diff
@@ -1,79 +1,148 @@
...
+    args = parser.parse_args()
```
FabienRoger, line 134: I mean this is wrong? No model name, no dataset, ... The train.sh script doesn't work.

FabienRoger:
> Hidden states are saved for all layers of the transformer by default, and probes are trained and evaluated for each layer of the network by default

I think it should be really easy and written in the README to only do this on the last layer and on every 4th layer. It makes ccs bit slower to run (especially when exploring hyperparameters), and it's heavy to store (30 layers x 4096 hidden dim for gpt j x 10000 samples x 4 bytes / float = 5GB per activation dataset, which is expensive and potentially slow to load).

elk/train.py
norabelrose: `train.sh` now no longer exists

PR #52, opened 2023-02-10
Make extract_hiddens usable programmatically

### LINKED ISSUE #48: Create Jupyter notebook-friendly API
Currently `extract_hiddens` takes an `argparse.Namespace` argument, which makes it awkward to use _programmatically_ or inside a notebook. We should do a bit of refactoring here to make sure the hidden state extraction and probe fitting can be called directly, outside of the CLI.

norabelrose:
Fixes #48


FabienRoger (approved):
lgtm

PR #53, opened 2023-02-10
Configure GitHub Actions CI For CPU Tests

### LINKED ISSUE #39: Create GitHub Actions CI Pipeline for CPU Tests
Since our aim is to have a mature codebase that can support many researchers onboarding, we need a way to catch bugs and breaking changes inadvertently introduced by contributors. Our ultimate aim will be to have a test suite that replicates results from previous experiments, we will also want unit tests to catch component-level issues.

While we work to configure GPU-based experiment tests, we can add quality controls quickly by setting up a PR pipeline for CPU-based unit tests. This pipeline can also ultimately run in parallel with a GPU-based CI pipeline and potentially catch issues faster.

I created a similar pipeline for GPT-NeoX. Tests can be marked as `cpu` and separately run. https://github.com/EleutherAI/gpt-neox/blob/main/.github/workflows/cpu_ci.yml

Kyle1668:
Closes #39

### DIFF
--- a/.github/workflows/cpu_ci.yml
+++ b/.github/workflows/cpu_ci.yml
@@ -31,4 +31,4 @@ jobs:
       #   run: python prepare_data.py
 
       - name: Run CPU Tests
-        run: pytest tests -m cpu
+        run: pytest -m cpu
--- a/.github/workflows/cpu_ci.yml
+++ b/.github/workflows/cpu_ci.yml
@@ -3,7 +3,7 @@ name: "Run CPU Tests"
 on: "push"
 
 jobs:
-  run-tests:
+  run-tests-python3_9:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


.github/workflows/cpu_ci.yml
```diff
@@ -0,0 +1,42 @@
+name: "Run CPU Tests"
```
Kyle1668, line 1: I've read that Actions has `composites` where we can make this more DRY. I don't think it's currently necessary with how simple this pipeline is.

norabelrose (approved):
Honestly it seems like you know more about GH Actions than I do so I'm content to merge this and see what happens. AFAICT there's no _simple_ way to test the actions before merging. If needed we can revert it.

Kyle1668:
@norabelrose Thanks for the review! The updated actions automatically run in the branch where these edits are made. See the `checks` tab of the PR.

image

I didn't in this case, but if I'm less certain about the changes I;m making and want a faster feedback loop, I use the [act](https://github.com/nektos/act) CLI tool. This tool runs the actions in a local Docker container and is pretty easy to use.

PR #55, opened 2023-02-12
Remove TrainParams

norabelrose:
Simplifies the training code a little bit, removing an internal dataclass

### DIFF
--- a/elk/training/ccs.py
+++ b/elk/training/ccs.py
@@ -1,6 +1,5 @@
 from .losses import ccs_squared_loss, js_loss
 from copy import deepcopy
-from dataclasses import dataclass
 from pathlib import Path
 from sklearn.metrics import roc_auc_score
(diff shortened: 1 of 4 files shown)


elk/training/ccs.py
```diff
@@ -141,31 +118,29 @@ def fit(
         if verbose:
```
lauritowal, line 118: Should we maybe remove verbose also from the css class, and just always print the things under the ifs ?

lauritowal (approved):

elk/training/ccs.py
norabelrose: Hmm I think we should either keep `verbose` or _never_ print, because I want to support programmatic uses where you may not want to print every time you call `fit`. Like imagine if PyTorch optimizers always printed when you initialized them or called `step` or something.

PR #56, opened 2023-02-12
Feature balance

lauritowal:


elk/extraction/dataset_preprocessing.py
```diff
@@ -0,0 +1,28 @@
...
+
+    subsets = []
+    for label in labels:
+        subsets.append(
+            dataset.filter(lambda x: x[label_column] == label).select(
+                range(min(counts))
+            )
+        )
+
+    return concatenate_datasets(subsets).shuffle()
```
norabelrose, line 28: We should probably support setting the `shuffle` seed here

elk/extraction/prompt_collator.py
```diff
@@ -48,8 +51,17 @@ def __init__(
...
         self.dataset = data[split]
+
+        if balance:
+            self.dataset = undersample(self.dataset, label_column)
+
         self.labels, counts = np.unique(self.dataset[label_column], return_counts=True)
         self.label_fracs = counts / counts.sum()
+
+        print(f"Class balance '{split}': {[f'{x:.2%}' for x in self.label_fracs]}")
+        if self.label_fracs[0] != self.label_fracs[1]:
```
norabelrose, line 62: This isn't robust in the case where `len(self.label_fracs) > 2`. You could instead do:
```
pivot, *rest = self.label_fracs
if not all(x == pivot for x in rest):
    print("Use arg --balance to force class balance")
```
It's also a little unclear if we want to print this for cases of _near_ class balance, like 51% to 49% or something. But to keep the code simple I'm fine with doing this weird floating point equality thing for now.

norabelrose (changes requested):
Almost good to go, just two small requests

norabelrose (approved):
LGTM

PR #57, opened 2023-02-13
Add recursive CCS

FabienRoger:


elk/training/ccs.py
```diff
@@ -102,7 +109,17 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
 
     def validate_data(self, data):
-        assert len(data) == 2 and data[0].shape == data[1].shape
+        assert (
+            len(data) == 2
+            and data[0].shape == data[1].shape
+            and data[0].dtype == data[1].dtype == self.dtype
+        ), "Data must be a tuple of two tensors of the same shape and dtype"
+
+    def correct_dtypes(
```
norabelrose, line 118: Why is this necessary at all? I don't see why we need to cast to a single dtype

elk/extensions/recursive_ccs/train.py
```diff
@@ -0,0 +1,110 @@
...
+from tqdm.auto import tqdm
+
+from elk.files import elk_cache_dir
+from elk.training.ccs import CCS
+from elk.training.preprocessing import load_hidden_states, normalize
+from elk.extensions.recursive_ccs.rccs import RecursiveCCS
+from elk.extensions.recursive_ccs.parser import get_rccs_training_parser
+
+
+@torch.autocast("cuda", enabled=False)
```
norabelrose, line 18: Should just be removed; there's no outer context that's turning `autocast` on so we don't need to explicitly turn it off here

elk/training/ccs.py
```diff
@@ -29,6 +30,7 @@ def __init__(
         loss: Literal["js", "squared"] = "squared",
         num_layers: int = 1,
         pre_ln: bool = False,
+        first_linear_parametrization: Optional[nn.Module] = None,
```
norabelrose, line 33: I would prefer to have `RecursiveCCS` just directly call `P.register_parametrization` on `.probe[0]` of the appropriate `CCS` object, instead of cluttering `CCS.__init__` with another parameter

elk/extensions/recursive_ccs/parser.py
```diff
@@ -0,0 +1,12 @@
+from elk.training.parser import get_training_parser
+from argparse import ArgumentParser
+
+
+def get_rccs_training_parser() -> ArgumentParser:
+    parser = get_training_parser()
+    parser.add_argument(
+        "--num-recursive-iterations",
```
norabelrose, line 8: Could be shortened to `--num-iterations` or `--num-iter` or `--num-directions`

elk/extensions/recursive_ccs/train.py
```diff
@@ -0,0 +1,110 @@
+import csv
```
norabelrose, line 1: It's not clear why this entire file is necessary. It seems to be mostly copied over from the primary `train.py`. Could we add a flag or a subparser to the main command instead?

norabelrose (changes requested):

elk/extensions/recursive_ccs/parser.py
FabienRoger: Done!

elk/training/ccs.py
FabienRoger: Done!

elk/extensions/recursive_ccs/train.py
FabienRoger: Done

FabienRoger: Also, torch's documentation advises against using autocast for training. Maybe this should be disabled in the regular train.py?

norabelrose: Um, that's just wrong? Autocast is frequently used for training. You can run into problems if you don't use `GradScaler` though

FabienRoger: But we don't use GradScaler, right?

elk/training/ccs.py
norabelrose: If you just call `.float()` on the input before calling `fit` this should work

elk/extensions/recursive_ccs/train.py
norabelrose: Yep that's why I'm getting rid of autocast in the `semi-supervised` branch. LBFGS currently isn't compatible with `GradScaler` annoyingly. In the future I might try to manually add grad scaling to LBFGS

PR #58, opened 2023-02-13
Initial implementation of semi-supervised training

### LINKED ISSUE #36: Support DDP and FSDP multi-gpu parallelism for hidden state extraction
Currently there's an option to "parallelize" the model, but this just uses _very_ naive pipeline parallelism with no prefetching, so it only saves VRAM and not wall clock time for running big models.

### LINKED ISSUE #54: Support semi-supervised probe training
Christiano et al. (2021) propose using _supervised_ learning to train a probe ("reporter") that answers questions about what a model knows, and using various types of regularizers to get it to generalize in the right way. This codebase currently follows the "regularizer-only" unsupervised approach taken in Burns et al. (2022), but I think we should support the whole continuum of possibilities here. It's quite possible that very small numbers of confidently labeled examples would make the approach a lot more robust.

norabelrose:
**Draft** PR, fixes #54. As a bonus, also partially fixes #36 (for DDP only)


FabienRoger (approved):
lgtm
(haven't tested it since I don't have multiple GPUs)

elk/extraction/prompt_collator.py
```diff
@@ -54,7 +53,10 @@ def __init__(
             raise ValueError(f"Dataset {path}/{name} has only one label")
         if max_examples:
             self.dataset = self.dataset.select(range(max_examples))
+        if dist.is_initialized():
+            self.dataset = self.dataset.shard(dist.get_world_size(), dist.get_rank())
 
+        self.dataset = self.dataset.shuffle(seed=seed)
```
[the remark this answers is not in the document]
norabelrose: It's weird but the HuggingFace docs explicitly state that you should shard _first_ and then shuffle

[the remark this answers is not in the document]
norabelrose: Sharding is deterministic btw

elk/extraction/extraction.py
```diff
@@ -9,7 +9,7 @@
 import torch.distributed as dist
 
 
-@torch.autocast("cuda", enabled=torch.cuda.is_available())
+@torch.autocast("cuda", enabled=torch.cuda.is_available())  # type: ignore
```
norabelrose, line 12: I'm not seeing typing issues here. Are you using mypy?

elk/training/ccs.py
```diff
@@ -140,58 +180,70 @@ def fit(
...
+
         # Calibrated accuracy
         cal_thresh = pred_probs.float().quantile(labels.float().mean())
-        cal_preds = pred_probs.gt(cal_thresh).squeeze(1).to(int)
-        raw_preds = pred_probs.gt(0.5).squeeze(1).to(int)
+        cal_preds = pred_probs.gt(cal_thresh).squeeze(1).to(torch.int)
+        raw_preds = pred_probs.gt(0.5).squeeze(1).to(torch.int)
 
-        auroc = float(roc_auc_score(labels.tolist(), pred_probs.tolist()))
+        auroc = 0.0  # float(roc_auc_score(labels.cpu(), pred_probs.cpu()))
```
norabelrose, line 209: Oh this needs to be reverted

elk/extraction/extraction.py
FabienRoger: I think this was a pylance error. I'll check

elk/training/ccs.py
norabelrose: Done

elk/extraction/extraction.py
FabienRoger: Pylance
Pytorch 1.13.1+cu117

elk/training/train.py
```diff
@@ -101,6 +98,10 @@ def train(args):
         if not args.skip_baseline and not dist.is_initialized():
             # TODO: Once we implement cross-validation for CCS, we should benchmark
             # against LogisticRegressionCV here.
+
+            train_labels_aug = torch.cat([train_labels, 1 - train_labels]).cpu()
```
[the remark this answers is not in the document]
FabienRoger: My bad, I had done sth wrong which causes it to be loaded elsewhere.

norabelrose:
> Another issue I just caught: shuffling now happens after selection of the first max_examples, which means that for datasets where points are sorted by label (such as imdb) you won't be able to use low number of max examples.

Fixed

elk/extraction/prompt_collator.py
```diff
@@ -52,12 +52,13 @@ def __init__(
...
         if len(self.labels) < 2:
             raise ValueError(f"Dataset {path}/{name} has only one label")
-        if max_examples:
-            self.dataset = self.dataset.select(range(max_examples))
         if dist.is_initialized():
             self.dataset = self.dataset.shard(dist.get_world_size(), dist.get_rank())
 
         self.dataset = self.dataset.shuffle(seed=seed)
+        if max_examples:
+            self.dataset = self.dataset.select(range(max_examples))
```
FabienRoger, line 60: Doesn't that select world_size x max_examples examples?

norabelrose: Yep!

elk/extraction/extraction.py
lauritowal: I don't see any issues here

elk/training/ccs.py
```diff
@@ -85,6 +90,9 @@ def reset_parameters(self):
             for layer in self.probe:
                 if isinstance(layer, nn.Linear):
                     layer.reset_parameters()
+        elif self.init == "zero":
```
lauritowal, line 93: Just curious, why did you add this?

elk/utils.py
```diff
@@ -1,4 +1,33 @@
+from torch import Tensor
+from torch.nn.parallel import DistributedDataParallel as DDP
 from typing import Callable, Mapping, TypeVar
+import torch.distributed as dist
+import torch.nn as nn
+
+
+def maybe_all_gather(x: Tensor) -> Tensor:
```
lauritowal, line 8: probably a dumb quesiton, but why the "maybe"s? Is it because you have maybe multiple processes or maybe just one?

lauritowal (approved):

elk/training/ccs.py
norabelrose: I noticed that you tend to get non-trivial probes with high accuracy even without any confidence loss, and I wanted to test the hypothesis that it's due to an initialization issue where it's hard for the optimizer to get to the trivial solution of outputting 0.5 all the time. Turns out to be true. See messages in the ELK channel a couple days ago

elk/utils.py
norabelrose: Yeah it's because these methods do nothing when the script isn't run with `torchrun`

PR #59, opened 2023-02-14
Use one file per layer

FabienRoger:


norabelrose:
I think this is going to be made obsolete by @AlexTMallen's PR once it's ready

PR #66, opened 2023-02-15
Rename CCS -> Reporter

### LINKED ISSUE #65: Rename "CCS" class to "Reporter"
In the original ELK paper, the probe used to answer questions using model hidden states is called a "reporter." I propose that we use this term to refer to probes in our library, since we plan to allow for general loss functions #17. We should probably do this _before_ the PyPI release to minimize breaking changes, especially because right now we save whole CCS modules directly with `pickle` instead of using state dicts.

norabelrose:
Fixes #65

### DIFF
--- a/elk/training/parser.py
+++ b/elk/training/parser.py
@@ -27,7 +27,7 @@ def add_train_args(parser: ArgumentParser):
         type=str,
         default="default",
         choices=("default", "spherical", "zero"),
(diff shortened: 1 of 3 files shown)


AlexTMallen (approved):
looks good!

lauritowal (approved):

PR #73, opened 2023-02-15
fixed max_example_bug

EricMungai97:
If the max_examples parameter is larger than the number of examples in the dataset, you will get an error when trying to select a range that is out of bounds. To fix this, you can check the number of examples in the dataset and adjust the max_examples parameter accordingly.


norabelrose (approved):
LGTM, thanks!

PR #74, opened 2023-02-16
Feature sweep

lauritowal:
Still WIP.


elk/evaluate/evaluate.py
```diff
@@ -0,0 +1,58 @@
+import csv
+
+import torch
+
+from elk.training.preprocessing import load_hidden_states, normalize
+
+
+def evaluate(args):
+    """
+    Note: eval is a reserved keyword in python, therefore we use evaluate instead
```
norabelrose, line 10: You can actually shadow the built in keyword if you want, it won't raise a SyntaxError

elk/evaluate/evaluate.py
```diff
@@ -0,0 +1,58 @@
...
+            "auroc",
+            "normalization",
+            "traind on dataset",
+            "eval on dataset",
+        ]
+
+        args.eval_dir.mkdir(parents=True, exist_ok=True)
+        with open(args.eval_dir / hidden_states_path, "w") as f:
+            writer = csv.writer(f)
+            writer.writerow(cols)
```
norabelrose, line 55: Sorry for having written it in this weird shitty way before, I think I did it to try to fix some `torch.distributed` issue. It would be better to move `writerow` up inside the `for hidden_state in layers` loop

norabelrose (changes requested):

elk/evaluate/evaluate.py
norabelrose: (Not saying this means you _should_ name it `eval`)

elk/__main__.py
```diff
@@ -1,12 +1,57 @@
...
+from elk.files import args_to_uuid, elk_cache_dir
+from elk.list import list_runs
+
+from .argparsers import get_extraction_parser, get_training_parser, get_evaluate_parser
+from .extraction.extraction_main import run as run_extraction
+from .training.train import train
+
+
+# TODO: Move function to a better place...
+def sweep(args):
```
lauritowal, line 18: Handling everything via the command line / args makes the sweep a bit annoying. I guess it might be better to have a .yaml file for the config... 

elk sweep burns.yaml

PR #78, opened 2023-02-16
Add "elk list" command

### LINKED ISSUE #77: Create an "elk list" command to show info about all cached runs
The command should show, at minimum:
- The model string
- The dataset(s)
- Creation date
- Filename / md5 hash
...in a table format.

norabelrose:
Fixes #77


norabelrose:
We might want to add a column that shows the disk space taken up by the hidden states

elk/__main__.py
```diff
@@ -41,8 +37,16 @@ def run():
...
     )
+    subparsers.add_parser("list", help="List all cached runs.")
     args = parser.parse_args()
 
+    # `elk list` is a special case
+    if args.command == "list":
+        list_runs(args)
+        return
+
+    from transformers import AutoConfig, PretrainedConfig
```
lauritowal, line 48: Is there a reason why this is here now?

lauritowal: I guess because of list_runs()

lauritowal (approved):

elk/__main__.py
norabelrose: It's to speed up imports

PR #83, opened 2023-02-17
Remove redundant 'elk train' command

norabelrose:
There's no context in which you would want to use `elk train` over `elk elicit`, and having the two commands around simply makes things more confusing.


elk/__main__.py
```diff
@@ -1,6 +1,6 @@
 """Main entry point for `elk`."""
 
-from .argparsers import get_extraction_parser, get_training_parser
+from .argparsers import add_train_args, get_extraction_parser
```
lauritowal, line 3: we probably want to rename add_train_args to add_elicit_args everywhere?

norabelrose: I think it's fine right now because `add_train_args` does add the arguments _relevant to training_. Arguably that whole function could be deleted and moved over into `__main__.py`, but I think I like the current layout slightly better

PR #82, opened 2023-02-17
(WIP) Use torch.multiprocessing to parallelize probe training across layers

anshradh:
Addresses #64

PR #86, opened 2023-02-18
Train reporters for different layers in parallel

### LINKED ISSUE #64: Parallelize probe training across layers
Instead of using `torch.distributed` to paralleize probe training, we can simply create a `multiprocessing.Queue` with one worker per GPU, and train the probes for each layer of a transformer in parallel.

norabelrose:
Fixes #64, based on draft PR #82


elk/training/train.py
```diff
@@ -54,114 +127,64 @@ def train(args):
...
-        with open(cache_dir / "eval.csv", "w") as f:
-            writer = csv.writer(f)
-            writer.writerow(cols)
+        writer = csv.writer(f)
+        writer.writerow(cols)
+
+        # Don't bother with multiprocessing if we're only using one device.
+        if args.num_devices == 1:
+            pbar = tqdm(iterator, total=L, unit="layer")
+            for i, (train_h, val_h) in enumerate(pbar):
```
AlexTMallen, line 157: Currently this claims that the layer is the same thing as the index into train_h and val_h and like you said this is broken. It might be good to flag that. This will be redone in versions where the layer information is available.

elk/training/train.py
```diff
@@ -11,25 +12,97 @@
...
-    the arguments passed in.
-    Consult parser.py for more details on the arguments.
+        stats += [lr_auroc, lr_acc]
+        with open(lr_dir / f"layer_{layer_index}.pkl", "wb") as file:
+            pickle.dump(lr_model, file)
 
-    Args:
-        args: The arguments for training.
-    """
+    with open(reporter_dir / f"layer_{layer_index}.pt", "wb") as file:
```
AlexTMallen, line 97: this saving convention in particular is what's confusing--"layer_{layer_index}.pkl" doesn't necessarily hold the data for layer `layer_index`

AlexTMallen (commented):
It looks good to me overall, except for the somewhat obscure handling of layer indices. It's not super critical to clarify that though since this is a temporary state of affairs while the layer information isn't available.

elk/training/train.py
norabelrose: Added a warning in `elk extract` when you try to save only a subset of layers as a bandaid solution

elk/training/train.py
norabelrose: Added a warning in `elk extract` when you try to save only a subset of layers as a bandaid solution

AlexTMallen (approved):
lgtm

elk/extraction/extraction_main.py
```diff
@@ -95,10 +102,14 @@ def extract(args, split: str):
     print("Loading datasets")
     silence_datasets_messages()
 
+    maybe_barrier()  # Not strictly necessary but makes the output cleaner
     extract(args, "train")
+    maybe_barrier()
+    print("wow")
```
lauritowal, line 108: I guess the wow thing shoudld be removed (?)

elk/extraction/extraction_main.py
```diff
@@ -95,10 +102,14 @@ def extract(args, split: str):
...
     extract(args, "validation")
 
-    if rank == 0:
+    print("hi")
```
lauritowal, line 111: and hi

elk/list.py
```diff
@@ -19,8 +19,11 @@ def list_runs(args):
     )
     for timestamp, run in subfolders:
         # Read the arguments used to run this experiment
-        with open(run / "args.json", "r") as f:
-            run_args = json.load(f)
+        try:
```
lauritowal, line 22: Isn't it cleaner to check for the file with an if instead of using a try / except and then continue?

norabelrose: The reason I did it this way is that `args.json` really _should_ exist, it's an anomaly if it doesn't exist. Arguably we should be indicating the error somewhere but `elk list` is going to change a lot soon anyway so I can't be bothered

elk/extraction/extraction_main.py
norabelrose: oh sorry that was a stupid debugging thing LOL

PR #89, opened 2023-02-18
Feature eval

lauritowal:
I created an own pull request from the eval which was part of the sweep pull request. 

(I am not evaluating LR, but we could add this in another PR, if needed)


elk/evaluation/evaluate.py
```diff
@@ -0,0 +1,65 @@
...
+            "cal_acc",
+            "auroc",
+            "normalization",
+            "name",
+            "hidden_states",
+        ]
+
+        transfer_eval = reporter_root_path / "transfer_eval"
+        transfer_eval.mkdir(parents=True, exist_ok=True)
+        with open(transfer_eval / f"{hidden_state_dir}.csv", "w") as f:
```
lauritowal, line 58: Need to probably clean up this part a bit.
Maybe, some additional refactoring is needed, to avoid having similar / duplicate code here and in the train.py

elk/__main__.py
```diff
@@ -36,8 +38,11 @@ def run():
     add_train_args(elicit_parser)
 
     subparsers.add_parser(
-        "eval", help="Evaluate a set of ELK reporters generated by `elk train`."
+        "eval",
+        help="Evaluate a set of ELK reporters" " generated by `elk train`.",
```
norabelrose, line 42: nit: this should be a single string literal, not two

elk/argparsers.py
```diff
@@ -210,3 +210,41 @@ def add_train_args(parser: ArgumentParser):
...
+    parser = ArgumentParser(add_help=False)
+    add_eval_args(parser)
+    return parser
+
+
+def add_eval_args(parser):
+    parser.add_argument(
+        "name",
+        type=str,
+        help="Name of the experiment containing" "the reporters you want to evaluate.",
```
norabelrose, line 225: Same problem here, and a few times below? Looks like you chose to break the literals into pieces much shorter than 88 characters, and then `black` came in and put them on the same line

elk/evaluation/evaluate.py
```diff
@@ -0,0 +1,63 @@
...
+
+                    result = reporter.score(
+                        (x0, x1),
+                        labels.to(args.device),
+                    )
+                    stats = [*result]
+                    stats += [args.normalization, args.name, hidden_state_dir]
+
+                    writer = csv.writer(f)
+                    if not csv_file.exists():
```
norabelrose, line 49: This is a little weird, I think I would just create a separate CSV for each reporter?

norabelrose (changes requested):
A couple nitpicks. We need to decide if we're going to merge this before or after changing how/where reporters are saved. I think I lean toward doing it after but idk

elk/evaluation/evaluate.py
lauritowal: No, this is only for creating the heading col once

elk/evaluation/evaluate.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+        assert len(set(labels)) > 1
+
+        _, hiddens = normalize(hiddens, hiddens, cfg.normalization)
+
+        reporter_root_path = elk_cache_dir() / cfg.source / "reporters"
+
+        reporters = []
+        for path in reporter_root_path.glob("*.pt"):
+            reporter = torch.load(path, map_location=cfg.device)
```
norabelrose, line 36: We shouldn't load all the reporters into memory at once, it's not necessary

elk/evaluation/evaluate.py
```diff
@@ -0,0 +1,112 @@
...
+        cfg.dump_yaml(f)
+
+    cols = ["layer", "loss", "acc", "cal_acc", "auroc"]
+    # Evaluate reporters for each layer in parallel
+    with mp.Pool(num_devices) as pool, open(out_dir / "eval.csv", "w") as f:
+        fn = partial(
+            evaluate_reporter, cfg, ds, devices=devices, world_size=num_devices
+        )
+        writer = csv.writer(f)
+        writer.writerow(cols)
```
norabelrose, line 106: Could you change this so that it mirrors how we're doing the progress bar and CSV writing in the latest commit?
Currently we use `imap_unordered` but then write to the CSV all at once inside a try-finally block to make sure it happens even when we crash

norabelrose (changes requested):

norabelrose (approved):
I don't like that there's a lot of duplicated code here, we should probably refactor to share more code with train.py but whatever

PR #92, opened 2023-02-20
Migrate CLI to `simple_parsing`, allowing for YAML based configs

norabelrose:
This PR migrates the CLI argument parsing and config management to the [SimpleParsing](https://github.com/lebrice/SimpleParsing) library.

Configs are specified using nested dataclasses. The top-level dataclass is `RunConfig`, which contains all information needed to train a reporter network, including the parameters used to extract hidden states from the model.


elk/extraction/prompt_dataset.py
```diff
@@ -1,4 +1,4 @@
-"""Collator for prompts."""
+"""prompts for prompts."""
```
AlexTMallen, line 1: Typo in comment here

AlexTMallen (approved):
Lgtm!

elk/extraction/prompt_dataset.py
norabelrose: Good catch, I did a stupid find and replace here

Issue #72, opened 2023-02-15
Add automatic type checking to the linting pipeline


norabelrose:
Right now we use type annotations but don't use any static analysis tools to check that the code is type-correct. We should look at using `mypy` or some other tool for this.

thejaminator:
@norabelrose Do you still need this? Given that pyright was added in  https://github.com/EleutherAI/elk/pull/90/

PR #93, opened 2023-02-21
Add mypy, fix types for mypy

thejaminator:
Closes https://github.com/EleutherAI/elk/issues/72

I understand that pyright is already added. If you guys choose to go ahead with pyright instead of mypy, this MR can be closed.


elk/extraction/prompt_dataset.py
```diff
@@ -89,8 +89,10 @@ def __init__(self, cfg: PromptConfig):
         self.strategy = cfg.strategy
 
         # TODO: Should we support IterableDataset?
-        ds_dict = load_dataset(*data_path)  # type: ignore
-        assert isinstance(ds_dict, DatasetDict)
+        _ds_dict = load_dataset(*data_path)  # type: ignore
+        assert isinstance(_ds_dict, DatasetDict)
+        ds_dict: DatasetDict = _ds_dict
```
thejaminator, line 94: unfortunately mypy needs this "dance"

elk/extraction/prompt_dataset.py
```diff
@@ -222,12 +224,12 @@ def __init__(self, cfg: PromptConfig):
...
                 )
 
-            self.fewshot_strata = [
+            self.fewshot_strata: list[Dataset] = [
                 ds_dict["train"].filter(lambda ex: ex[label_column] == i)
                 for i in range(self.num_classes)
             ]
         else:
-            self.fewshot_strata: list[Dataset] = []
+            self.fewshot_strata = []
```
thejaminator, line 231: otherwise mypy will complain that the type already got declared. so the first branch must be the first the declare it.

elk/training/reporter.py
```diff
@@ -325,15 +331,17 @@ def train_loop_adam(
...
             optimizer.zero_grad()
 
             loss = self.loss(probe(x0), probe(x1), labels)
             loss.backward()
             optimizer.step()
 
-        return float(loss)
+        return (
+            float(loss) if loss is not None else raise_assertion_error("Loss is None")
+        )
```
thejaminator, line 344: i'm raising an assertion error here, otherwise we can keep the original implementation of returning torch.inf if num_epochs is somehow 0. I think its better to raise earlier

mypy.ini
```diff
@@ -0,0 +1,24 @@
...
+ignore_missing_imports = True
+
+[mypy-transformers.*]
+ignore_missing_imports = True
+
+[mypy-pynvml.*]
+ignore_missing_imports = True
+
+[mypy-promptsource.*]
+ignore_missing_imports = True
```
thejaminator: config for mypy

elk/training/reporter.py
norabelrose: We should probably just raise before the for-loop even starts if `num_epochs < 1`. Would that fix the mypy error?

elk/extraction/prompt_dataset.py
norabelrose: Yep this seems sufficiently ugly that I'm not sure I want it in the code. Is there really no cleaner way of doing the same thing?

norabelrose:
Thanks for the PR, I think we definitely do want to stick with Pyright as our _primary_ type checker but it would be a bonus if the repo also type checks under Mypy. I just have mild code style concerns about having to re-declare variables in order to narrow the type. If there's some way around that then I'd be happy to merge this PR I think.

norabelrose:
@thejaminator Maybe we should just introduce a wrapper around `typing.cast` that also runtime-asserts the right type? That would actually be cleaner than what we're doing right now in a lot of places (sticking an assert statement right after a declaration). If it's supported by both Mypy and Pyright that'd be great.

elk/extraction/prompt_dataset.py
thejaminator: see https://github.com/EleutherAI/elk/pull/93/files/245859cb3d5a74e569e0800c191ee7ef4a02feb7..51db6995fe583d29ae5e3a863fda75724af6ce76#r1113128336

elk/extraction/prompt_dataset.py
thejaminator: > @thejaminator Maybe we should just introduce a wrapper around `typing.cast` that also runtime-asserts the right type? That would actually be cleaner than what we're doing right now in a lot of places (sticking an assert statement right after a declaration). If it's supported by both Mypy and Pyright that'd be great.

ah yeah, basically the assert_is_instance i implement here https://github.com/EleutherAI/elk/pull/93/files/245859cb3d5a74e569e0800c191ee7ef4a02feb7..51db6995fe583d29ae5e3a863fda75724af6ce76#r1113128336

elk/training/reporter.py
thejaminator: i don't think i can be fixed. because there is no "type-system" thing that says num_epochs will be a positive integer >= 1. so even if you raise before that, you can't prove to the type system that "this will definitely loop one time".

IDK why pyright doesn't complain though.

thejaminator: ok actually i'll try it out by reverting to torch.inf

elk/training/losses.py
```diff
@@ -1,25 +1,36 @@
...
+    return _js_loss(logit0, logit1, confidence=0.0, base=2.0)
+
+
+def _js_loss(
+    logit0: Tensor,
+    logit1: Tensor,
+    confidence: float,
+    base: float,
+) -> Tensor:
     p0, neg_p1 = logit0.sigmoid(), 1 - logit1.sigmoid()
```
thejaminator, line 34: fix so that js_loss has the exact same type signature

thejaminator:
> Thanks for the PR, I think we definitely do want to stick with Pyright as our _primary_ type checker but it would be a bonus if the repo also type checks under Mypy. I just have mild code style concerns about having to re-declare variables in order to narrow the type. If there's some way around that then I'd be happy to merge this PR I think.

I would hesitate against then making mypy in the pipeline because it adds more load for PR creators to having both pyright and mypy happy :D

PR #99, opened 2023-03-01
Fix single gpu stat logging

thejaminator:
Seems like we were forgetting to log the stats if the user selected to use a gpu.

So the csv output would be empty with only the columns e.g.
```
layer,train_loss,loss,acc,cal_acc,auroc,lr_auroc,lr_acc
```

### DIFF
--- a/elk/training/train.py
+++ b/elk/training/train.py
@@ -171,7 +171,7 @@ def train(cfg: RunConfig, out_dir: Optional[Path]):
         if len(devices) == 1:
             pbar = tqdm(iterator, total=L, unit="layer")
             for i, (train_h, val_h) in enumerate(pbar):
-                train_reporter(
+                layer_stat = train_reporter(
                     cfg,
                     i,
                     train_h,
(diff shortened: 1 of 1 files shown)


elk/training/train.py
```diff
@@ -181,6 +181,7 @@ def train(cfg: RunConfig, out_dir: Optional[Path]):
                     device=devices[0],
                     out_dir=out_dir,
                 )
+                writer.writerow([L - i, *layer_stat])
```
thejaminator, line 184: following the layer numbering of multigpu

elk/training/train.py
```diff
@@ -181,6 +181,7 @@ def train(cfg: RunConfig, out_dir: Optional[Path]):
                     device=devices[0],
                     out_dir=out_dir,
                 )
+                writer.writerow([L - i] + [f"{s:.4f}" for s in layer_stat])
```
thejaminator, line 184: i'll like to refactor this reporter functions so that they give back something like a `EvalCSVResult` eventually.

```
from elk.training.reporter import EvalResult


class EvalCSVResult:
    layer: int
    train_loss: float
    eval_result: EvalResult
    # Only available if reporting baseline
    lr_auroc: Optional[float]
    # Only available if reporting baseline
    lr_acc: Optional[float]
```

and then another function does the responsbility of writing to a CSV / rounding to 4 d.p
but i'll do that in another MR :)

AlexWan0: Sounds good, could you make a new issue for this?

AlexWan0:
Lgtm, thanks for the PR!

AlexWan0 (approved):

PR #100, opened 2023-03-03
Move promptsource into a submodule; remove vestigial license

norabelrose:
We need to move the `promptsource` fork into a submodule so that it gets distributed jointly inside the PyPI package. I also went ahead and removed the second MIT license, with a Collin Burns copyright, from LICENSE.md because we now use literally no code from the original codebase.


.github/workflows/cpu_ci.yml
```diff
@@ -7,6 +7,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          submodules: recursive
```
thejaminator, line 12: needs to checkout with the submodule

elk/extraction/prompt_dataset.py
```diff
@@ -3,7 +3,7 @@
 from dataclasses import dataclass
 from datasets import DatasetDict, load_dataset
 from numpy.typing import NDArray
-from promptsource.templates import DatasetTemplates
+from promptsource_module.promptsource.templates import DatasetTemplates
```
thejaminator, line 6: i think this was the main :feelsgood: 
basically locally we had `promptsource` installed. so the tests /typecheck would be okay.
On the pipeline, instead of `promptsource.templates`, we actually needed to call `promptsource.promptsource.templates`

anyways i've renamed it creatively to `promptsource_module` to save us from potential future headaches while we still use the submodule

pyproject.toml
```diff
@@ -36,8 +34,10 @@ dependencies = [
     "tqdm>=4.0.0",
     # 4.0 introduced the breaking change of using return_dict=True by default
     "transformers>=4.0.0",
+    # required by promptsource, which doesn't specify a version
+    "jinja2"
```
thejaminator, line 38: we need to explicitly specify this now since we don't install it from the fork

README.md
```diff
@@ -42,6 +42,15 @@ Currently, our code doesn't quite support distributed training of the probe. Run
...
 ## Development
+To clone the repo and its submodules
+```bash
+git clone --recurse-submodules https://github.com/EleutherAI/elk.git
+```
+If you already cloned the repo but are missing the promptsource_module submodule, run
+```bash
+git submodule update --init --recursive
+```
+
```
thejaminator, line 53: added a readme here

PR #101, opened 2023-03-04
Remove promptsource submodule; move essential bits into repo proper

norabelrose:
We ended up only needing the `templates/` folder with all the YAML files and (most of) `templates.py`


norabelrose:
This PR also bumps the version to 0.1.1 since this is sort of a major internal refactor that changes what packages get installed

AlexTMallen (approved):
lgtm! Perhaps we should credit the promptsource repo either at the top of `templates.py` or in a new file inside `promptsource/`? Currently I don't think there's any link to their repo.

PR #103, opened 2023-03-05
Fit logistic regression models on the GPU

### LINKED ISSUE #88: Use `Reporter` for the supervised baseline, not sklearn `LogisticRegression`
Currently the main bottleneck in the training phase is fitting the scikit-learn `LogisticRegression` models, not the actual fitting of the reporters. This is because `LogisticRegression` is done on the CPU. But we could easily replicate ~the same results by fitting `Reporter` objects with a purely supervised loss, and this should speed up the training process dramatically.

norabelrose:
Fixes #88 
Note: merge #101 first

In `main`, _most_ of the wall clock time used up by the training process consists of training the logistic regression baseline models with scikit-learn. This is because scikit-learn runs on the CPU only. In this PR, I replace `sklearn.LogisticRegression` with a simple PyTorch module called `Classifier`, which uses LBFGS to fit its coefficients to a batch of data. **This can speed up training by an order of magnitude**.

Since `Classifier` is so fast, I also use it to _predict the pseudo-labels_ (e.g. "Yes", "No") for each contrast pair, and measure its AUROC. This "pseudo-AUROC" is a potentially useful diagnostic for determining how well the normalization is working. If the pseudo-AUROC is high, that's a bad sign: it means the pseudo-labels are linearly separable and the unsupervised algorithm may pick up on that more than anything else. Currently the code prints a warning whenever the pseudo-AUROC is over 0.6, but we might want to tune this or make it customizable in the future. On the small number of (model, dataset) pairs I've looked at, though, it seems like the pseudo-AUROC is very close to 0.5, so I think the naive approach mostly works.

Also, I switched the training script from using `Pool.imap_unordered` to normal `Pool.imap`, which ensures that the entries in the eval CSV are sorted (making it much easier to read at a glance). The downside is that the progress bar can be "jumpier." We might be able to do something better in the future— the stupidly simple idea would be to stick with `imap_unordered` and then only print the eval CSV at the end, but I kinda wanted to avoid that since esp. during debugging the training script might crash and we want to make sure partial results get written to disk. IDK.


FabienRoger:
I really like measuring pseudo-AUROC.
Have you seen it go above 0.5 in some cases?

FabienRoger:
Using Pool.imap looks good to me. I guess that it doesn't make the bar too jumpy if the number of tasks is larger enough relative to the number of workers, with is hopefully the regime most of us are working with?

elk/training/classifier.py
```diff
@@ -0,0 +1,64 @@
...
+    """Linear classifier trained with supervised learning."""
+
+    def __init__(
+        self, input_dim: int, num_classes: int = 1, device: Optional[str] = None
+    ):
+        super().__init__()
+
+        self.linear = torch.nn.Linear(input_dim, num_classes, device=device)
+        self.linear.bias.data.zero_()
+        self.linear.weight.data.zero_()
```
FabienRoger, line 19: this made me slightly worried at first, but I guess it's always fine since its output is never used as input to another layer?

elk/training/classifier.py
```diff
@@ -0,0 +1,64 @@
...
+
+        optimizer.step(closure)
+        return float(loss)
+
+    def nullspace_project(self, x: torch.Tensor) -> torch.Tensor:
+        """Project the given data onto the nullspace of the classifier."""
+
+        # https://en.wikipedia.org/wiki/Projection_(linear_algebra)
+        A = self.linear.weight.data.T
+        P = A @ torch.linalg.solve(A.mT @ A, A.mT)
```
FabienRoger, line 63: You use that because it's more accurate than using gram Schmidt then projecting in the usual way? Looks good to me.
(Note: you drop the gradient, I hope this won't cause trouble to users wanting to do something weird).

FabienRoger (approved):
LGTM
(Haven't looked into the promptsource integration)

elk/training/classifier.py
norabelrose: Oh, I didn't even think about orthogonalizing `A` and then projecting. I just saw this formula in the Wikipedia article and it seems to work 😅

And yeah, I think including the gradient could cause more bugs than anything? But idk. I was primarily thinking of using this for [INLP](https://arxiv.org/abs/2004.07667). The idea is you could project the data onto the nullspace of the classifier for the pseudo-labels as a more robust form of normalization. But it seems that at least on the datasets I've looked at this isn't actually necessary (the pseudo-labels aren't linearly separable after subtracting the mean).

elk/training/classifier.py
norabelrose: Oh I was doing this because I think logistic regression models are usually initialized to zero? It _shouldn't_ matter since this is always a convex problem, although maybe in weird cases it could matter due to the 1e-4 stopping condition

lauritowal (approved):
Should work now, I've solved the merge conflicts and committed some small additional changes @norabelrose

PR #105, opened 2023-03-06
Bug/more than one variant

AlexTMallen:
When the `num_variants` dimension of the data input to training was greater than 1, the logistic regression and metrics wouldn't work properly. I reshaped and broadcast the tensors to fix this issue.

### DIFF
--- a/elk/training/reporter.py
+++ b/elk/training/reporter.py
@@ -121,7 +121,10 @@ class Reporter(nn.Module):
 
         self.init = cfg.init
         self.device = device
-        self.unsupervised_loss = js_loss if cfg.loss == "js" else ccs_squared_loss
+        self.unsupervised_loss = {
+            "js": js_loss, 
+            "squared": ccs_squared_loss, 
+        }[cfg.loss]
         self.supervised_weight = cfg.supervised_weight
 
     def reset_parameters(self):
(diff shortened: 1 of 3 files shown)


elk/training/reporter.py
```diff
@@ -296,9 +299,10 @@ def score(
         cal_preds = pred_probs.gt(cal_thresh).squeeze(1).to(torch.int)
         raw_preds = pred_probs.gt(0.5).squeeze(1).to(torch.int)
 
-        auroc = float(roc_auc_score(labels.cpu(), pred_probs.cpu()))
-        cal_acc = cal_preds.eq(labels.reshape(-1)).float().mean()
-        raw_acc = raw_preds.eq(labels.reshape(-1)).float().mean()
+        tiled_labels = labels.repeat_interleave(pred_probs.shape[1])
```
norabelrose, line 302: We should probably have some sort of a comment explaining what this is doing? Right now it's very opaque

elk/training/train.py
```diff
@@ -101,8 +101,14 @@ def train_reporter(
     stats = [layer, train_loss, *val_result]
 
     if not cfg.skip_baseline:
-        train_labels_aug = torch.cat([train_labels, 1 - train_labels]).cpu()
-        val_labels_aug = torch.cat([val_labels, 1 - val_labels]).cpu()
+        train_labels_aug = (
```
norabelrose, line 104: Ditto here

norabelrose (changes requested):
Just add some comments and I'll merge

norabelrose (approved):

PR #107, opened 2023-03-07
Prompt var loss

AlexTMallen:
Adds a prompt variance loss term. Fixes a tensor reshaping bug that snuck back into main for num_variants > 1.


norabelrose (approved):
LGTM

PR #111, opened 2023-03-07
Custom loss

AlexTMallen:
adds custom loss.
Examples:
`--loss consistency_squared 0.2*confidence_squared 2*prompt_var_squared`
`--loss ccs`  (this refers to the original CCS loss, equivalent to `--loss consistency_squared confidence_squared`)

If there are any disagreements about naming in the registry before the PR that might be good to resolve.


elk/training/losses.py
```diff
@@ -71,8 +149,6 @@ def prompt_var_loss(logit0: Tensor, logit1: Tensor) -> Tensor:
             "Only one variant provided. Prompt variance loss will equal CCS loss."
         )
     p0, p1 = logit0.sigmoid(), logit1.sigmoid()
-    consistency = p0.sub(1 - p1).square().mean()
-    confidence = torch.min(p0, p1).square().mean()
     mean_p0, mean_p1 = p0.mean(dim=-1, keepdim=True), p1.mean(dim=-1, keepdim=True)
```
norabelrose, line 152: you could actually just use `p0.var(dim=-1, keepdim=True)` here I think

norabelrose (approved):
Good job

PR #117, opened 2023-03-08
PCA init MVP

norabelrose:
`"pca"` is now a possible value for `ReporterConfig.init`

### DIFF
--- a/elk/training/reporter.py
+++ b/elk/training/reporter.py
@@ -55,7 +55,7 @@ class ReporterConfig(Serializable):
     activation: Literal["gelu", "relu", "swish"] = "gelu"
     bias: bool = True
     hidden_size: Optional[int] = None
-    init: Literal["default", "spherical", "zero"] = "default"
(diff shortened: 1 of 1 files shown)


elk/training/reporter.py
```diff
@@ -268,9 +269,15 @@ def fit(
         best_state: dict[str, torch.Tensor] = {}  # State dict of the best run
         x0, x1 = contrast_pair
 
-        for _ in range(cfg.num_tries):
+        for i in range(cfg.num_tries):
             self.reset_parameters()
 
+            # This is sort of inefficient but whatever
```
AlexTMallen, line 275: So you put the PCA init outside the reset_parameters() call because it needs access to the index i? IIUK, a new probe is trained for each of the top `num_tries` PCs. In future PRs it might be good to clean this up a bit and add explanation (and like you mentioned not recomputing the top i principal components every time).

AlexTMallen (approved):
lgtm as an mvp

PR #118, opened 2023-03-09
Add test for the classifier to be roughly the same as sklearn

thejaminator:


tests/test_classifier.py
```diff
@@ -0,0 +1,30 @@
...
+        n_samples=1000, n_features=input_dims, random_state=0
+    )
+    # use float32 for the features so it's the same as the default dtype for torch
+    features = _features.astype(np.float32)
+    truths = _truths.astype(np.float32)
+    # train a logistic regression model on the data. No regularization
+    model = LogisticRegression(penalty="none", solver="lbfgs")
+    model.fit(features, truths)
+    # train a classifier on the data
+    classifier = Classifier(input_dim=input_dims, device="cpu")
```
thejaminator, line 24: classifier doesn't expose the dtype used. by default the torch linear layer is float32.
 so we'll turn the sklearn features from float64 to float32

norabelrose (approved):
LGTM thanks!

PR #120, opened 2023-03-10
Added error message for prompt-based loss and num_variants=1

Benw8888:
The loss functions ccs_prompt_var and prompt_var_squared cause NaN errors when --num_variants is set to 1, because of the variance calculation.

I added an error message in train.py for this.

As ccs_prompt_var and num_variants=1 are set by default, this means an error is thrown by default, which is not ideal.

### DIFF
--- a/elk/training/losses.py
+++ b/elk/training/losses.py
@@ -146,7 +146,7 @@ def prompt_var_loss(logit0: Tensor, logit1: Tensor, coef: float = 1.0) -> Tensor
     assert len(logit0.shape) in [1, 2]
     if logit0.shape[-1] == 1:
         warnings.warn(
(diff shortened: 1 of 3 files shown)


norabelrose (approved):
LGTM, thanks

AlexTMallen:
Ah it looks like torch.var by default computes the sample variance, which divides by n-1, rather than n. The previous code was dividing by n so returned the equivalent of CCS loss. But your solution seems fine too, since we don't really care about being able to run prompt_var with num_variants=1

norabelrose:
@AlexTMallen yeah you can also do `unbiased=False` but I think it's better to just error probably

PR #121, opened 2023-03-10
Support boolq inference

AlexTMallen:
boolq has a boolean type label column, so I added support for that. I also added jinja templates for boolq from Sanh et. al and a few of my own.


norabelrose (approved):
LGTM

PR #122, opened 2023-03-11
Sort outputs by variant id during extraction

AlexTMallen:
This enables all the hiddens in a particular "row" (index into the `num_variants` axis) to be of the same variant when using all available templates.

Note that when `num_variants` is less than the number of available templates, it will continue to be the case that different subsets of the templates will be used for each example.


AlexTMallen:
It turns out that the quality of trained LR and VINC models depends sensitively on the shuffling of the prompts--I haven't yet totally figured out what's going on but I have verified that the only difference between the two cases was the data ordering.

AlexTMallen:
This sensitivity turns out to be explained by the bug fixed in #127

norabelrose (approved):
LGTM

PR #127, opened 2023-03-15
Fix: PromptDataset inserting fake labels into its output

AlexTMallen:
This bug affects all previous experiments with `num_variants>1`.


norabelrose (approved):
LGTM

PR #128, opened 2023-03-15
Remove marker for cpu test, add marker for future gpu tests

thejaminator:
This PR removes the pytest.mark.cpu decorators, so all tests are automatically added to the CI. 

For future tests that require the GPU, they can be marked with `pytest.mark.gpu` instead. These tests won't run on the cpu CI

### DIFF
--- a/.github/workflows/cpu_ci.yml
+++ b/.github/workflows/cpu_ci.yml
@@ -21,8 +21,8 @@ jobs:
       - name: Type Checking
         uses: jakebailey/pyright-action@v1
 
-      - name: Run CPU Tests
-        run: pytest -m cpu
+      - name: Run normal tests, excluding GPU tests
+        run: pytest -m "not gpu"
 
   run-tests-python3_10:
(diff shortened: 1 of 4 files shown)


.github/workflows/cpu_ci.yml
```diff
@@ -42,5 +42,5 @@ jobs:
       - name: Type Checking
         uses: jakebailey/pyright-action@v1
 
-      - name: Run CPU Tests
-        run: pytest -m cpu
+      - name: Run normal tests, excluding GPU tests
+        run: pytest -m "not gpu"
```
thejaminator, line 46: in a further PR i can do something like

```
    strategy:
      matrix:
        python-versions: [3.9, "3.10"]
        os: [ubuntu-18.04, macos-latest]
    runs-on: ${{ matrix.os }}
```
so that we don't have to keep remembering to edit these two lines

norabelrose (approved):
LGTM, thanks!

PR #133, opened 2023-03-16
Replace held_out_split with select_train_val_splits

norabelrose:
This PR factors out the logic for determining which splits of a dataset should be used for training & validation into a function in `data_utils.py` called `select_train_val_splits`. This replaces the older `held_out_split` function which was less robust.

### DIFF
--- a/elk/utils/data_utils.py
+++ b/elk/utils/data_utils.py
@@ -54,7 +54,7 @@ def select_train_val_splits(raw_splits: Iterable[str]) -> tuple[str, str]:
         Split.TEST: 2,
     }
     splits = sorted(raw_splits, key=lambda k: priorities.get(k, 100))  # type: ignore
-    assert len(splits) >= 2, "Must have train and val/test splits"
+    assert len(splits) >= 2, "Must have at least two of train, val, and test splits"
 
     return tuple(splits[:2])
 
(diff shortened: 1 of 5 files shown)


elk/utils/data_utils.py
```diff
@@ -39,14 +46,17 @@ def get_columns_all_equal(dataset: DatasetDict) -> list[str]:
...
-        raise ValueError("No validation or test split found")
+def select_train_val_splits(raw_splits: Iterable[str]) -> tuple[str, str]:
+    """Return splits to use for train and validation, given an Iterable of splits."""
+    priorities = {
+        Split.TRAIN: 0,
+        Split.VALIDATION: 1,
+        Split.TEST: 2,
+    }
+    splits = sorted(raw_splits, key=lambda k: priorities.get(k, 100))  # type: ignore
+    assert len(splits) >= 2, "Must have train and val/test splits"
```
AlexTMallen: This error message is no longer true: should say "must have at least two of train, val, and test splits".

AlexTMallen (approved):
LGTM

elk/utils/data_utils.py
norabelrose: yeah

PR #134, opened 2023-03-16
Misleading prompts + zero-shot baseline comparison

AlexWan0:
This PR adds support for misleading prompt prefixes, specified in `extraction/misleading_prefixes.yaml`:

E.g., to use the prompt in Figure 5 of the original paper, run:
```bash
elk elicit allenai/unifiedqa-t5-3b imdb --mislead_type Prefix --prefix_name default
```

This also adds zero-shot evaluation as a comparison point. To get zero-shot results, run:
```bash
elk eval zero-shot allenai/unifiedqa-t5-3b imdb
```

Still need to run some tests to make sure zero-shot scores are reasonable.


norabelrose:
Too many merge conflicts to be worth salvaging this PR, but we should try to incorporate the misleading prompt prefixes soon

PR #135, opened 2023-03-17
Support L2 regularization & cross validation for Classifier

norabelrose:
This PR adds an `l2_penalty` parameter to `Classifier.fit` with a default value of 0.1— _this is a change_ from the previous behavior, where there was no penalty by default.

I initially tried to exactly imitate the behavior of scikit-learn's `C` inverse regularization parameter, but I couldn't quite figure out how they're computing the final loss. Based on their code it seems like they're doing some weird thing where they're summing the BCE loss over the samples rather than taking the average and this changes the scale of everything, making it dependent on the number of samples. But that didn't seem to give the exact same results either, so idk. I gave up on exactly imitating it— the tests only check that when `l2_penalty` is set to 0.0, the results are ~the same.

This PR also adds a relatively well optimized `fit_cv` method that uses warm-starting to get at least a 2x speed up over a naive approach where you start from a zero initialization every time. I initially wanted to parallelize this code over the folds but this seemed like it would be a real pain in the ass that would complicate the code substantially, and I'm not sure there would be a significant speed boost at the end of the day (at least not without rewriting PyTorch's LBFGS optimizer which I don't want to do right now).

There is a question of whether we want to use `fit_cv` _by default_ in `train.py`. I think we probably should, but it does get us back into the territory where `Classifier` is taking up more compute than the actual VINC algorithm. At the moment the code does use `fit_cv` and doesn't actually give you an option to turn this off (which should probably be changed).

### DIFF
--- a/elk/training/classifier.py
+++ b/elk/training/classifier.py
@@ -173,17 +173,6 @@ class Classifier(torch.nn.Module):
         mean_losses = losses.mean(dim=0)
         best_idx = mean_losses.argmin()
 
-        # Check the health of the regularization path- we expect a U shape
-        lhs, rhs = mean_losses[:best_idx], mean_losses[best_idx:]
-        if not len(lhs) or not len(rhs):
-            warnings.warn("The best penalty is at the edge of the regularization path.")
-        else:
-            lhs_monotonic_decreasing = lhs.diff().le(0).all()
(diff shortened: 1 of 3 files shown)


lauritowal (approved):

AlexTMallen (approved):
LGTM

PR #137, opened 2023-03-17
Custom PyTorch implementation of eigsh

norabelrose:
Introduces a custom PyTorch implementation of the Lanczos method for computing the top k eigenpairs of a symmetric matrix, `lanczos_eigsh`. This replaces the SciPy implementation and the naive dense eigendecomposition method that were previously used in `EigenReporter`.

This method can be several times faster than `torch.linalg.eigh` for large matrices, and is quite accurate and stable. It also supports warm-starting.

### DIFF
--- a/elk/eigsh.py
+++ b/elk/eigsh.py
@@ -1,6 +1,7 @@
 from torch import Tensor
 from typing import Literal, Optional
 import torch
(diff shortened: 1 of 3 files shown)


AlexTMallen (approved):
LGTM

PR #139, opened 2023-03-18
Go back to using 1000 datapoints as the default

norabelrose:
This PR also changes the default `num_variants` to `-1`, meaning we use all the prompt templates by default.


AlexTMallen (approved):
LGTM

PR #144, opened 2023-03-22
Set --balance to True by default

norabelrose:
This is the most sensible default; `False` is almost always not what you want


derpyplops (approved):

lauritowal (approved):

norabelrose:
Irrelevant now

PR #147, opened 2023-03-23
Refactor train eval

lauritowal:


derpyplops:
Looks like type tests are failing. The type of device should probably be str

elk/evaluation/evaluate.py
```diff
@@ -1,121 +1,77 @@
...
 
 @dataclass
-class EvaluateConfig(Serializable):
-    target: ExtractionConfig
+class Eval(Serializable):
+    """
+    Full specification of a reporter evaluation run.
+
+    Args:
+        data: Config specifying hidden states on which the reporter will be trained.
```
norabelrose, line 20: Do you mean the hidden states on which it will be _**evaluated**_?

elk/utils/typing.py
```diff
@@ -10,3 +12,23 @@ def assert_type(typ: Type[T], obj: Any) -> T:
...
+    """Converts float32 to float16, then reinterprets as int16."""
+    return x.type(torch.float16).view(torch.int16)
+
+
+def int16_to_float32(x: torch.Tensor) -> torch.Tensor:
+    """Converts int16 to float16, then reinterprets as float32."""
+    return x.view(torch.float16).type(torch.float32)
+
+
+def upcast_hiddens(hiddens: torch.Tensor) -> torch.Tensor:
```
norabelrose, line 27: This function definitely isn't needed, right? You're just calling `int16_to_float32` and _redundantly_ asserting the type of the hiddens (which already have to be a `torch.Tensor` since you specified that in the type signature of the function)

elk/utils/data_utils.py
```diff
@@ -131,14 +133,14 @@ def undersample(
...
 
-def float32_to_int16(x: torch.Tensor) -> torch.Tensor:
-    """Converts float32 to float16, then reinterprets as int16."""
-    return x.type(torch.float16).view(torch.int16)
-
-
-def int16_to_float32(x: torch.Tensor) -> torch.Tensor:
-    """Converts int16 to float16, then reinterprets as float32."""
-    return x.view(torch.float16).type(torch.float32)
+def get_layers(ds: DatasetDict) -> List[int]:
```
norabelrose, line 136: Yeah this should have been a function before, thanks

elk/utils/data_utils.py
```diff
@@ -49,13 +48,16 @@ def get_columns_all_equal(dataset: DatasetDict) -> list[str]:
     return pivot
 
 
-def select_train_val_splits(raw_splits: Iterable[str]) -> tuple[str, str]:
-    """Return splits to use for train and validation, given an Iterable of splits."""
-    priorities = {
+def select_train_val_splits(
+    raw_splits: Iterable[str],
+    priorities: dict = {
```
norabelrose, line 53: I don't think the priorities should be customizable here. This also violates the usual rule that you shouldn't have _mutable_ objects as default values for function arguments, because really weird stuff can happen

elk/training/train.py
```diff
@@ -52,90 +39,47 @@ class RunConfig(Serializable):
...
-    lr_dir = out_dir / "lr_models"
-    reporter_dir = out_dir / "reporters"
-
-    lr_dir.mkdir(parents=True, exist_ok=True)
-    reporter_dir.mkdir(parents=True, exist_ok=True)
-    stats = [layer, pseudo_auroc, train_loss, *val_result]
-
-    if not cfg.skip_baseline:
+@dataclass
+class TrainRun(Run):
```
norabelrose, line 70: I'm wondering if this could just be called `Train`

elk/run.py
```diff
@@ -0,0 +1,111 @@
...
+            train_split, val_split = select_train_val_splits(
+                self.dataset, priorities=priorities
+            )
+            train, val = self.dataset[train_split], self.dataset[val_split]
+
+            train_labels = cast(Tensor, train["label"])
+            val_labels = cast(Tensor, val["label"])
+
+            train_h, val_h = normalize(
+                upcast_hiddens(train[f"hidden_{layer}"]),  # type: ignore
```
norabelrose, line 77: Instead of type: ignore we should use `assert_type`

elk/run.py
```diff
@@ -0,0 +1,111 @@
...
+    ) -> tuple:
+        """Prepare the data for training and validation."""
+
+        with self.dataset.formatted_as("torch", device=device, dtype=torch.int16):
+            train_split, val_split = select_train_val_splits(
+                self.dataset, priorities=priorities
+            )
+            train, val = self.dataset[train_split], self.dataset[val_split]
+
+            train_labels = cast(Tensor, train["label"])
```
norabelrose, line 73: We should also use `assert_type` here. Basically the rule is to never use `cast`.

Internally `assert_type` does call `cast`, but it also does a runtime check that the type is in fact what you're saying it is, while `cast` just takes your word for it.

elk/files.py
```diff
@@ -36,3 +40,43 @@ def memorably_named_dir(parent: Path):
...
     out_dir.mkdir(parents=True, exist_ok=True)
     return out_dir
+
+
+def create_output_directory(
+    out_dir: Optional[Path] = None, default_root_dir: Path = elk_reporter_dir()
+) -> Path:
+    """Creates an output directory"""
+    if out_dir is None:
+        # default_root_dir.mkdir(parents=True, exist_ok=True)
```
norabelrose, line 50: Remove this commented out code

norabelrose (changes requested):
Primarily stylistic changes needed

elk/utils/data_utils.py
```diff
@@ -131,14 +132,14 @@ def undersample(
...
-
-
-def int16_to_float32(x: torch.Tensor) -> torch.Tensor:
-    """Converts int16 to float16, then reinterprets as float32."""
-    return x.view(torch.float16).type(torch.float32)
+def get_layers(ds: DatasetDict) -> List[int]:
+    """Get a list of indices of hidden layers given a `DatasetDict`."""
+    layers = [
+        int(feat[len("hidden_") :])
+        for feat in ds["train"].features
```
AlexTMallen, line 139: ds might not contain a "train" split, so we should just take ds.values().pop().features or the equivalent.

elk/evaluation/evaluate.py
```diff
@@ -1,121 +1,78 @@
...
-        for feat in ds["train"].features
-        if feat.startswith("hidden_")
-    ]
+    def evaluate_reporter(self, layer: int, devices: list[str], world_size: int):
+        """Evaluate a single reporter on a single layer."""
+        device = self.get_device(devices, world_size)
 
-    devices = select_usable_devices(cfg.num_gpus)
-    num_devices = len(devices)
+        _, _, test_x0, test_x1, _, test_labels = self.prepare_data(
```
AlexTMallen, line 55: I think we should split up the extraction of train and test data so we don't have to do this

AlexTMallen (changes requested):
I think this looks good overall! I just want to make the train/val change for `Eval`. I can write the code for this.

elk/utils/data_utils.py
lauritowal: don't you always still get at least train if there is no split?

thejaminator:
^ sorry my bad, was branching off from you and forgot to switch my branch when pushing. corrected it and ignore the previous push

elk/evaluation/evaluate.py
```diff
@@ -1,121 +1,78 @@
...
+            which contains the reporters directory.
+        normalization: The normalization method to use. Defaults to "meanonly". See
+            `elk.training.preprocessing.normalize()` for details.
+        num_gpus: The number of GPUs to use. Defaults to -1, which means
+            "use all available GPUs".
+        debug: When in debug mode, a useful log file is saved to the memorably-named
+            output directory. Defaults to False.
+    """
+
+    data: Extract
```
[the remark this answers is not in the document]
lauritowal: Just for the others from our conversation in the chat:
> Your change requests should probably be maybe implemented as a new feature by itself, since the current pull request is only about refactoring, but we can add them soonish after merging...

elk/utils/data_utils.py
AlexTMallen: I'm thinking about the refactor I want to do to support evaluations on a single split, and in the case of E.g. truthful QA there is only a validation split.

AlexTMallen (approved):
lgtm. We can support single splits in the next pr.

PR #148, opened 2023-03-23
`select_usable_gpus` will now wait for gpus to have min memory available

AlexTMallen:
This should reduce OOMs caused by multiple people using the same GPU.


norabelrose (changes requested):

norabelrose (approved):
LGTM

PR #149, opened 2023-03-23
Smoke tests with tiny gpt2, fix CCSReporter

thejaminator:
This MR adds a smoke test for CCSReporter with tinygpt2

Subsequent issues:
- Fix EigenReporter for tinygpt2
- Make the tests work for tiny deberta (or some other MLM?)

### DIFF
--- a/elk/training/ccs_reporter.py
+++ b/elk/training/ccs_reporter.py
@@ -195,10 +195,11 @@ class CcsReporter(Reporter):
 
             alpha = self.config.supervised_weight
             preds = p0.add(1 - p1).mul(0.5).squeeze(-1)
-            # unsqueeze and broadcast to match the shape of preds
-            # TODO: not sure what to do here actually
-            labels_unsqueezed = labels.unsqueeze(-1).expand_as(preds)
-            bce_loss = bce(preds, labels_unsqueezed.type_as(preds))
+            # broadcast the labels, and flatten the predictions
+            # so that both are 1D tensors
+            broadcast_labels = labels.repeat_interleave(preds.shape[1]).float()
+            flattened_preds = preds.cpu().flatten()
+            bce_loss = bce(flattened_preds, broadcast_labels.type_as(flattened_preds))
             loss = alpha * bce_loss + (1 - alpha) * loss
 
         elif self.config.supervised_weight > 0:
--- a/elk/training/ccs_reporter.py
+++ b/elk/training/ccs_reporter.py
@@ -195,9 +195,10 @@ class CcsReporter(Reporter):
 
             alpha = self.config.supervised_weight
             preds = p0.add(1 - p1).mul(0.5).squeeze(-1)
-            # TODO: wrong dims on CCS?
-            sum_preds = preds.sum(-1).sigmoid()
-            bce_loss = bce(sum_preds, labels.type_as(sum_preds))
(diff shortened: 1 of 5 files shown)


elk/training/ccs_reporter.py
```diff
@@ -195,7 +195,11 @@ def loss(
 
             alpha = self.config.supervised_weight
             preds = p0.add(1 - p1).mul(0.5).squeeze(-1)
-            bce_loss = bce(preds, labels.type_as(preds))
+            # broadcast the labels, and flatten the predictions
+            # so that both are 1D tensors
+            broadcast_labels = labels.repeat_interleave(preds.shape[1]).float()
+            flattened_preds = preds.cpu().flatten()
+            bce_loss = bce(flattened_preds, broadcast_labels.type_as(flattened_preds))
             loss = alpha * bce_loss + (1 - alpha) * loss
```
thejaminator, line 203: @norabelrose @AlexTMallen had to fix this bug to get the test to pass

tests/test_smoke_elicit.py
```diff
@@ -0,0 +1,66 @@
...
+from elk.extraction import PromptConfig
+from elk.training import CcsReporterConfig, EigenReporterConfig
+from elk.training.train import train, RunConfig
+
+"""
+TODO: These tests should work with deberta
+but you'll need to make deberta fp32 instead of fp16
+because pytorch cpu doesn't support fp16
+"""
+
```
thejaminator, line 15: basically the deberta model itself does some fp16 operations that aren't supported. but tiny-gpt doesn't. so that why tiny-gpt works. 

i'll try to figure that out in another MR. i think we could convert fp16 models to fp32 if its CPU, but seems hacky?

tests/test_smoke_elicit.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    Currently this test fails with
+    u -= torch.einsum("...ij,...i->...j", V[..., :k, :], proj)
+    V[..., k, :] = F.normalize(u, dim=-1)
+    ~~~~~~~~~ <--- HERE
+
+    u[:] = torch.einsum("...ij,...j->...i", A, V[..., k, :])
+
+    RuntimeError: select(): index 1 out of range for tensor of size [1, 2]
+    at dimension 0
```
thejaminator, line 48: EigenReporterConfig is broken for tiny-gpt2. didn't investigate, would like to push the fix for CCS first

tests/test_smoke_elicit.py
norabelrose: I actually don't view this as hacky. In general float16 isn't supported on CPUs. We could check the whether we're running on CPU right when we call `AutoModel.from_pretrained` and tell HF to always load the model as float32 in that case, and use the dtype of the checkpoint otherwise.

tests/test_smoke_elicit.py
norabelrose: Thanks for noticing this, I just pushed a fix

tests/test_smoke_elicit.py
norabelrose: ^ just made this change, so if you want to add a deberta test you can. I think tiny gpt2 is probably sufficient for now though

norabelrose (approved):
After my changes it LGTM

PR #154, opened 2023-03-25
Replace flake with ruff

thejaminator:


.pre-commit-config.yaml
```diff
@@ -14,11 +14,11 @@ repos:
...
     -   id: black
--   repo: https://github.com/pycqa/flake8
-    rev: '6.0.0'
+-   repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: 'v0.0.257'
     hooks:
-    -   id: flake8
-        args: ["--ignore=E203,F401,W503", --max-line-length=88]
+        - id: ruff
+          args: [--fix, --exit-non-zero-on-fix]
```
thejaminator, line 21: this makes ruff automatically remove unused imports

elk/training/__init__.py
```diff
@@ -1,4 +1,12 @@
...
+all = [
+    "Reporter",
+    "ReporterConfig",
+    "CcsReporter",
+    "CcsReporterConfig",
+    "EigenReporter",
+    "EigenReporterConfig",
+    "RunConfig",
+    "OptimConfig",
+]
```
thejaminator, line 12: > this makes ruff automatically remove unused imports

but that also means that you need to explicitly export imports

.pre-commit-config.yaml
thejaminator: see the remove imports here https://github.com/EleutherAI/elk/pull/154/commits/f337f745225e93c714b018bf600d18d757c5779b

elk/training/__init__.py
norabelrose: Oh this is a bit annoying actually...

ChatGPT suggests that you can automate this like so:
```
# Automatically fill the __all__ field
__all__ = [name for name, value in globals().items() if not name.startswith('__') and callable(value)]
```
appending this to the end of the `__init__` files

thejaminator: ruff itself (and the user's own linters) wouldn't be able to understand that. That you've reeexported everything inside.
So whoever uses the elk pacakge will still get warnings, and ruff itself will warn us too. (False positive but annoying for everyone)


I've added 
```
ignore-init-module-imports = true
```
to the pyproject.toml. So what ruff will do is to tell us that we've imported something in `__init__.py` but didn't add it to`__all__`. That should fix future mistakes when we forget to declare it.

elk/extraction/prompt_loading.py
```diff
@@ -9,13 +9,9 @@
...
 from datasets import (
-    interleave_datasets,
     load_dataset,
-    ClassLabel,
     Dataset,
     Features,
-    IterableDataset,
-    Sequence,
 )
 from datasets.distributed import split_dataset_by_node
```
thejaminator, line 16: unused imports

norabelrose (approved):
LGTM

Issue #153, opened 2023-03-25
train_reporter should return a dataclass of stats instead of a list


thejaminator:
What stats `train_reporter` returns can be pretty confusing. We can return a dataclass instead so that its clearer

e.g.
```py
from elk.training.reporter import EvalResult

class EvalCSVResult:
    layer: int
    train_loss: float
    eval_result: EvalResult
    # Only available if reporting baseline
    lr_auroc: Optional[float]
    # Only available if reporting baseline
    lr_acc: Optional[float]
```

PR #155, opened 2023-03-26
Create dataclasses for writing to CSV, refactor CSV logging, fix eval CSV columns

thejaminator:
Depends on https://github.com/EleutherAI/elk/pull/147
Closes https://github.com/EleutherAI/elk/issues/153

### DIFF
--- a/elk/evaluation/evaluate.py
+++ b/elk/evaluation/evaluate.py
@@ -10,7 +10,7 @@ from elk.extraction.extraction import Extract
 from elk.files import elk_reporter_dir
 from elk.run import Run
 from elk.training import Reporter
-from elk.training.train_result import EvalLog
+from elk.evaluation.evaluate_log import EvalLog
 from elk.utils import select_usable_devices
 
 
--- a/elk/evaluation/evaluate.py
+++ b/elk/evaluation/evaluate.py
@@ -41,8 +41,6 @@ class Eval(Serializable):
 
     def execute(self):
         transfer_eval = elk_reporter_dir() / self.source / "transfer_eval"
-        # TODO: Why is this different?
-        # cols = ["layer", "loss", "acc", "cal_acc", "auroc"]
 
         run = Evaluate(cfg=self, out_dir=transfer_eval)
         run.evaluate()
--- a/elk/evaluation/evaluate.py
+++ b/elk/evaluation/evaluate.py
@@ -5,10 +5,10 @@ from typing import Literal, Optional
 import torch
 from simple_parsing import Serializable, field
 
-from datasets import Split
 from elk.extraction.extraction import Extract
 from elk.files import elk_reporter_dir
 from elk.run import Run
+from elk.training.train_result import EvalStatResult
 
 
 @dataclass
@@ -38,9 +38,10 @@ class Eval(Serializable):
 
     def execute(self):
         transfer_eval = elk_reporter_dir() / self.source / "transfer_eval"
-        cols = ["layer", "loss", "acc", "cal_acc", "auroc"]
+        # TODO: Why is this different?
+        # cols = ["layer", "loss", "acc", "cal_acc", "auroc"]
 
-        run = Evaluate(cfg=self, eval_headers=cols, out_dir=transfer_eval)
+        run = Evaluate(cfg=self, out_dir=transfer_eval)
         run.evaluate()
 
 
@@ -48,7 +49,9 @@ class Eval(Serializable):
 class Evaluate(Run):
     cfg: Eval
 
-    def evaluate_reporter(self, layer: int, devices: list[str], world_size: int):
+    def apply_to_single_layer(
+        self, layer: int, devices: list[str], world_size: int = 1
+    ) -> EvalStatResult:
         """Evaluate a single reporter on a single layer."""
         device = self.get_device(devices, world_size)
 
@@ -74,5 +77,4 @@ class Evaluate(Run):
 
     def evaluate(self):
         """Evaluate the reporter on all layers."""
-
-        self.apply_to_layers(func=self.evaluate_reporter)
+        self.apply_to_layers()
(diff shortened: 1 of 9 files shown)


elk/evaluation/evaluate_log.py
```diff
@@ -0,0 +1,27 @@
...
+    def csv_columns() -> list[str]:
+        return ["layer", "acc", "cal_acc", "auroc", "ece"]
+
+    def to_csv_line(self) -> list[str]:
+        items = [
+            self.layer,
+            self.eval_result.acc,
+            self.eval_result.cal_acc,
+            self.eval_result.auroc,
+            self.eval_result.ece,
```
thejaminator, line 23: previously the columns were ["layer", "loss", "acc", "cal_acc", "auroc"] for whatever Eval logged. So i think was a bug where acc was written to to "loss" instead, and "cal_acc" to "acc" instead, etc.

tests/test_write_iterator_to_file.py
```diff
@@ -0,0 +1,168 @@
...
+                out_dir=tmp_path,
+                csv_columns=csv_columns,
+                to_csv_line=to_csv_line,
+            )
+    except ValueError:
+        # We expect a ValueError to be raised,
+        # and don't want to fail the test
+        pass
+    # We should still have results for layer 1, 3, even though layer 2 failed
+    with open(tmp_path / "eval.csv", "r") as f:
```
thejaminator, line 161: test for kabooms

elk/utils/csv.py
```diff
@@ -0,0 +1,39 @@
...
+        for row in iterator:
+            row_buf.append(row)
+    finally:
+        # Make sure the CSV is written even if we crash or get interrupted
+        sorted_by_layer = sorted(row_buf, key=lambda x: x.layer)
+        for row in sorted_by_layer:
+            row = to_csv_line(row)
+            writer.writerow(row)
+        if debug:
+            save_debug_log(dataset, out_dir)
```
thejaminator, line 39: factored this out so its much easier to test

elk/utils/csv.py
```diff
@@ -0,0 +1,39 @@
...
+
+from elk.evaluation.evaluate_log import EvalLog
+from elk.logging import save_debug_log
+from elk.training.train_log import ElicitLog
+
+"""A generic log type that contains a layer field
+The layer field is used to sort the logs by layer."""
+Log = TypeVar("Log", EvalLog, ElicitLog)
+
+
```
thejaminator, line 15: this typevar has to be bounded on `EvalLog, ElicitLog` so you can sort by the layer properly with pyright

elk/run.py
```diff
@@ -88,28 +90,36 @@ def prepare_data(
...
-
-        devices = select_usable_devices(self.cfg.num_gpus)
-        num_devices = len(devices)
+    def apply_to_layers(
+        self,
+        func: Callable[[int], Log],
+        num_devices: int,
+        to_csv_line: Callable[[Log], list[str]],
+        csv_columns: list[str],
+    ):
```
thejaminator, line 99: i made func have a signature like
```
Callable[[int], Log]
```

previously it was something like

```
Callable[[int, int, list[int]], Log] # layer, device, list of devices
```

which could be pretty confusing

elk/run.py
```diff
@@ -88,28 +90,36 @@ def prepare_data(
...
+
+        Args:
+            func: The function to apply to each layer.
+                The int is the index of the layer.
+            num_devices: The number of devices to use.
+            to_csv_line: A function that converts a Log to a list of strings.
+                This has to be injected in because the Run class does not know
+                the extra options e.g. skip_baseline to apply to function.
+            csv_columns: The columns of the CSV file."""
         self.out_dir = assert_type(Path, self.out_dir)
```
thejaminator, line 111: i made func have a signature like
```
Callable[[int], Log]
```

previously it was something like

```
Callable[[int, int, list[int]], Log] # layer, device, list of devices
```

which could be pretty confusing

thejaminator: previously func wasn't annotated so pyright wouldn't catch keyword mispelling / mistakes in `partial`. 

we could also make it a paramspec instead so you indiate that is a function that has keywords world_size, devices, but i didn't want to complicate things

elk/evaluation/evaluate.py
```diff
@@ -38,17 +41,18 @@ class Eval(Serializable):
...
 
 
 @dataclass
 class Evaluate(Run):
     cfg: Eval
 
-    def evaluate_reporter(self, layer: int, devices: list[str], world_size: int):
+    def evaluate_reporter(
+        self, layer: int, devices: list[str], world_size: int = 1
+    ) -> EvalLog:
```
thejaminator, line 55: it now returns a dataclass instead of a list[str]. 
Which i think is nicer since
- Whoever wants to call Evaluate manually (e.g. in a notebook) gets back a dataclass rather than a list of unlabelled strings
- The responsbility of writing a CSV format is done somewhere else.

elk/run.py
```diff
@@ -88,28 +90,38 @@ def prepare_data(
...
+        # Should we write to different CSV files for elicit vs eval?
```
lauritowal, line 112: > Should we write to different CSV files for elicit vs eval?

we are writing to different CSV files, since for eval the out_dir is usually by default under ~/transfer_eval... However, we could the name "eval.csv" into something like "transfer_eval.csv", maybe.

elk/run.py
```diff
@@ -88,28 +90,38 @@ def prepare_data(
...
-            try:
-                for i, *stats in tqdm(mapper(fn, layers), total=len(layers)):
-                    row_buf.append([i] + [f"{s:.4f}" for s in stats])
-            finally:
-                # Make sure the CSV is written even if we crash or get interrupted
-                for row in sorted(row_buf):
-                    writer.writerow(row)
-                if self.cfg.debug:
-                    save_debug_log(self.dataset, self.out_dir)
+            iterator: Iterator[Log] = tqdm(  # type: ignore
```
lauritowal, line 116: Is the type: ignore necessary here?

lauritowal (approved):
Thanks for the improvements!  Looks good to me, I've added a few comments, but I'll merge it into the refactor-train-eval branch later.

PR #157, opened 2023-03-26
Gather eval stats

reaganjlee:
Adds in a new function ``get_metric_across_layers`` that can be used to get a metric score of a reporter across layers automatically. 
Some questions/things to work on
1. I'm looking to make this callable from CLI with something similar to ``elk gather`` (open to thoughts on the command)
2. The function requires that there is a reporter directory to run the tests. I test it by having a reporter directory and setting ENV_DIR to that, but am unsure on how to test this with pytest on the library by itself


elk/evaluation/gather.py
```diff
@@ -0,0 +1,56 @@
...
+        ``reporter_name``: Name of 'model_name--dataset_name' to get data from
+        ``reporter_dir``: Name of the directory to look for reporter data
+        ``save_csv``: Boolean to optionally save output as a csv file
+
+    Outputs:
+        ``values``: A pandas series that has an index of layer_num and value of metric
+    """
+    REPORTER_DIR = os.path.abspath(reporter_dir)
+
+    if not os.path.exists(REPORTER_DIR):
```
lauritowal, line 26: Why do you need to create the reporter_dir first?

elk/evaluation/gather.py
```diff
@@ -0,0 +1,56 @@
+import pandas as pd
+import numpy as np
+import matplotlib.pyplot as plt
```
lauritowal, line 3: there are missing imports

elk/promptsource/templates.py
```diff
@@ -24,7 +24,7 @@
 
 # These are users whose datasets should be included in the results returned by
 # filter_english_datasets (regardless of their metadata)
-INCLUDED_USERS = {"Zaid", "craffel", "lauritowal"}
+INCLUDED_USERS = {"Zaid", "craffel", "lauritowal", "christykoh"}
```
lauritowal, line 27: You should remove the template changes, since this is not related to visualizing the stats

lauritowal (changes requested):
Thanks for your pull request! I've added some comments

norabelrose:
Not clear what this is doing; happy to look at an updated version in the future

PR #165, opened 2023-04-04
Store final layer LM output and record AUROC and acc

norabelrose:
Also refactors the logging to use Pandas, and removes Python 3.9 support because I was too lazy to refactor my cool span conversion function to work around 3.9

### DIFF
--- a/elk/extraction/extraction.py
+++ b/elk/extraction/extraction.py
@@ -322,10 +322,6 @@ def extract(cfg: "Extract", num_gpus: int = -1) -> DatasetDict:
 
     mp.set_start_method("spawn", force=True)  # type: ignore[attr-defined]
 
-    import multiprocess as mp
-
-    mp.set_start_method("spawn")  # type: ignore[attr-defined]
-
     ds = dict()
     for split, builder in builders.items():
         builder.download_and_prepare(num_proc=len(devices))


elk/extraction/extraction.py
```diff
@@ -148,24 +135,69 @@ def extract_hiddens(
...
+                # Compute the log probability of the answer tokens if available
+                if type(outputs).__name__.startswith("CausalLMOutput"):
+                    start, end = convert_span(
+                        offsets, (answer_start, answer_start + len(choice["answer"]))
+                    )
+                    log_p = outputs.logits[..., start - 1 : end - 1, :].log_softmax(
+                        dim=-1
+                    )
+                    tokens = inputs.input_ids[..., start:end, None]
+                    lm_preds[i, j] = log_p.gather(-1, tokens).sum()
```
lauritowal, line 193: This part is still a bit confusing to me, and I'll have a look at it in detail tomorrow. Apart from that, the code seems to work correctly and everything else lgtm.

ss-waree (commented):
I tested and it seems that the LM outputs appear when I used elicit but not when I use eval -- not sure if I'm missing something. 

Edit: nevermind this isn't necessary

elk/extraction/extraction.py
```diff
@@ -148,24 +135,69 @@ def extract_hiddens(
...
 
+                # TODO: Do something smarter than "rindex" here. Really we want to
+                # get the span of the answer directly from Jinja, but that doesn't
+                # seem possible. This approach may fail for complex templates.
+                answer_start = text.rindex(choice["answer"])
+
+                # Only feed question, not the answer, to the encoder for enc-dec models
+                if model.config.is_encoder_decoder:
+                    # TODO: Maybe make this more generic for complex templates?
+                    text = text[:answer_start].rstrip()
```
AlexTMallen, line 163: Why do we strip the right here? This kind of assumes the answer string starts with whitespace correct?

norabelrose: No, we don't assume that. It's not the case that the concatenation of the encoder input and the decoder input have to like, make sense or be a full sentence or something. I'm sort of assuming that the answer string does _not_ start with whitespace. And the encoder is not going to expect trailing whitespace in its input (although I doubt it actually matters)

elk/extraction/extraction.py
```diff
@@ -100,32 +105,16 @@ def extract_hiddens(
...
-    # If this is an encoder-decoder model we don't need to run the decoder at all.
-    # Just strip it off, making the problem equivalent to a regular encoder-only model.
-    if is_enc_dec:
-        # This isn't actually *guaranteed* by HF, but it's true for all existing models
-        if not hasattr(model, "get_encoder") or not callable(model.get_encoder):
-            raise ValueError(
-                "Encoder-decoder model doesn't have expected get_encoder() method"
-            )
-
-        model = assert_type(PreTrainedModel, model.get_encoder())
```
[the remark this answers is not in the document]
norabelrose: Yes because that complicates things and is not how you actually do inference on encoder-decoder models normally

elk/extraction/extraction.py
```diff
@@ -148,24 +135,69 @@ def extract_hiddens(
...
-            for j in range(2):
-                text = record[j]["text"]
+            for j, choice in enumerate(record):
+                text = choice["text"]
                 variant_inputs.append(text)
 
+                # TODO: Do something smarter than "rindex" here. Really we want to
+                # get the span of the answer directly from Jinja, but that doesn't
+                # seem possible. This approach may fail for complex templates.
+                answer_start = text.rindex(choice["answer"])
```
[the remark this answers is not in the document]
norabelrose: I hope it's not necessary to include the answer in the question, because that would slow down inference (since if the answer only appears at the end you can cache the keys and values).

> I'm wondering if we should store the completion string (the second half of the jinja template) in the prompt dataset, and if it's empty and the user is attempting to use encoder-decoder models, we raise an exception.

Yes.

In general I'm like, yes, we should do the templates better, but I don't want that to hold up merging this PR

elk/extraction/extraction.py
norabelrose: Is it still confusing

elk/utils/data_utils.py
```diff
@@ -132,11 +132,12 @@ def binarize(template: Template, label: int, new_label: int, rng: Random) -> Tem
     true = answer_choices[label]
     false = rng.choice([c for c in answer_choices if c != true])
 
-    assert new_label in (0, 1)
+    # What order are we going to present the answer choices in?
+    new_label = rng.choice([0, 1])
```
AlexTMallen, line 136: This won't work because all variants for a given example must have the same new_label

norabelrose: sigh

norabelrose: Technically we _could_ give all variants different labels but let's not do that

norabelrose: I reluctantly just reverted this— really we need to destroy this and replace it with something actually correct

AlexTMallen (approved):
I reviewed all the changes since my last review and they look good to me

PR #166, opened 2023-04-04
check is streamable

AlexTMallen:
This also removes hard-coded checks for datasets


elk/utils/data_utils.py
```diff
@@ -120,3 +120,15 @@ def binarize(template: Template, label: int, new_label: int, rng: Random) -> Tem
     )
 
     return new_template
+
+
+def is_streamable(ds: IterableDataset, max_examples: int) -> bool:
```
norabelrose, line 125: oh is this actually fast? I guess this could be okay... I was imagining something where like, `BalancedSampler` internally checks if it's skipped over too many samples without yielding anything, maybe throwing `StopIteration` or whatever in that case, and then we pick up on that

But idk maybe this is just too complex

norabelrose:
> he hanging is caused by it streaming 750 examples to check if they all have the same label. I found that streaming the data necessarily takes this much time (in my case, about 8 seconds). This should only impact performance when it throws an Exception for the dataset not being streamable.

It's possible that I'm misremembering but I do distinctly remember actually going through with a print statement to see what was going on and it like, stayed frozen even after it had hit the appropriate number of examples

AlexTMallen:
>  It's possible that I'm misremembering but I do distinctly remember actually going through with a print statement to see what was going on and it like, stayed frozen even after it had hit the appropriate number of examples

I can't replicate this when I use print statements. If you're using `max_examples / world_size` less than 100, it checks the first 100 streaming examples as a minimum. As soon as it reaches the appropriate iteration it raises an exception for me.

norabelrose:
We don't even support streaming datasets anymore

PR #167, opened 2023-04-05
Added confidence, invariance, consistency metrics

ss-waree:


elk/training/eigen_reporter.py
```diff
@@ -105,6 +105,18 @@ def contrastive_xcov(self) -> Tensor:
     def intercluster_cov(self) -> Tensor:
         return self.intercluster_cov_M2 / self.n
 
+    @property
+    def confidence(self) -> Tensor:
+        return self.weight.mT @ self.intercluster_cov @ self.weight
+
+    @property
+    def invariance(self) -> Tensor:
+        return self.weight.mT @ self.intracluster_cov @ self.weight
```
norabelrose, line 114: Sorry can you add a negative sign to this? Right now it's _higher_ when there's _more_ variation across prompts (so less invariance).

The idea is that, for each of these metrics, it should be "higher is better" roughly.

elk/training/eigen_reporter.py
```diff
@@ -105,6 +105,18 @@ def contrastive_xcov(self) -> Tensor:
...
+
+    @property
+    def consistency(self) -> Tensor:
```
norabelrose, line 117: Same here

norabelrose (changes requested):
Just fix the signs

norabelrose (approved):
LGTM! Thanks

PR #168, opened 2023-04-05
Turn on import sorting

norabelrose:
Enable isort (`I` code) in `ruff`, enforcing PEP-whatever import sorting


thejaminator (approved):
Lgtm!

PR #177, opened 2023-04-08
Force spawn start method

norabelrose:
For some bizarre reason this has started to be necessary now


Kyle1668 (approved):
I confirmed that this resolves the issue

PR #181, opened 2023-04-12
Add portuguese templates for boolq, ag_news, imdb

ChristyKoh:


elk/promptsource/templates.py
```diff
@@ -24,7 +24,7 @@
 
 # These are users whose datasets should be included in the results returned by
 # filter_english_datasets (regardless of their metadata)
-INCLUDED_USERS = {"Zaid", "craffel", "lauritowal"}
+INCLUDED_USERS = {"Zaid", "craffel", "lauritowal", "christykoh"}
```
lauritowal, line 27: Not sure if this is still necessary (?). If yes, we maybe want to create a user for elk library itself on hugginface to add all the needed datasets, so we don't need to keep adding users @ChristyKoh

PR #182, opened 2023-04-12
Prevent invalidation of the hidden state cache when num_gpus changes

### LINKED ISSUE #102: Hidden state cache shouldn't be invalidated when we're using different devices
Currently the precise device strings returned by `select_usable_devices` are hashed into the fingerprint of the hidden state dataset. This causes caching to fail in a lot of cases where it we should be loading the hidden states from cache. We should figure out some way to make sure the device strings aren't hashed into the dataset fingerprint.

norabelrose:
The fix here is moderately hacky; I had to override `create_config_id` in `_GeneratorConfig` to manually remove any trace of the number of GPUs from the generator kwargs before they get hashed.

Fixes #102.

### DIFF
--- a/elk/extraction/extraction.py
+++ b/elk/extraction/extraction.py
@@ -1,6 +1,7 @@
 """Functions for extracting the hidden states of a model."""
 import logging
 import os
+from copy import copy
 from dataclasses import InitVar, dataclass
 from itertools import islice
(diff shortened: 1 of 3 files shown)


AlexTMallen (approved):
LGTM--this works on the UW machine as well

PR #187, opened 2023-04-13
add custom_ds_n CLI arg

reaganjlee:
Allows ``--custom_ds_n [custom_name]`` to shorten long naming from multiple datasets  

Unmodified, the output path to eliciting with datasets ``super_glue boolq`` and ``christykoh/boolq_pt`` 
``elk-reporters/bigscience/bloomz-7b1-mt/christykoh/boolq_pt, super_glue boolq/quizzical-mccarthy``

To cd to this reporter requires, ``elk-reporters/bigscience/bloomz-7b1-mt/christykoh/"boolq_pt, super_glue boolq"/quizzical-mccarthy`` where path is divided with arbitrary number of ``/`` in the datasets as opposed to where datasets are actually divided 

You can use this keyword when running ``elk elicit``/``eval`` without typing the full path to group the experiments by datasets, such as in the example above, ``--custom_ds_n multiboolq`` to get ``elk-reporters/bigscience/bloomz-7b1-mt/multiboolq/quizzical-mccarthy``

### DIFF
--- a/elk/run.py
+++ b/elk/run.py
@@ -35,7 +35,7 @@ class Run(ABC):
     cfg: Union["Elicit", "Eval"]
     out_dir: Optional[Path] = None
     dataset: DatasetDict = field(init=False)
-    custom_ds_n: str = None
+    custom_ds_n: Optional[str] = None
 
     def __post_init__(self):
         # Extract the hidden states first if necessary
@@ -45,7 +45,6 @@ class Run(ABC):
             # Save in a memorably-named directory inside of
             # ELK_REPORTER_DIR/<model_name>/<dataset_name>
             ds_name = ", ".join(self.cfg.data.prompts.datasets)
-            print(f"self.custom_root_name: {self.custom_ds_n}")
             if self.custom_ds_n:
                 ds_name = self.custom_ds_n
             root = elk_reporter_dir() / self.cfg.data.model / ds_name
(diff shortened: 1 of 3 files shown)


elk/run.py
```diff
@@ -35,6 +35,7 @@ class Run(ABC):
     cfg: Union["Elicit", "Eval"]
     out_dir: Optional[Path] = None
     dataset: DatasetDict = field(init=False)
+    custom_ds_n: Optional[Path] = None
```
lauritowal, line 38: > random unimportant thing: because we now require Python >= 3.10, we can replace Union[X, Y] with X | Y and Optional[X] with X | None in type annotations

elk/run.py
```diff
@@ -44,6 +45,8 @@ def __post_init__(self):
             # Save in a memorably-named directory inside of
             # ELK_REPORTER_DIR/<model_name>/<dataset_name>
             ds_name = ", ".join(self.cfg.data.prompts.datasets)
+            if self.custom_ds_n:
+                ds_name = self.custom_ds_n
             root = elk_reporter_dir() / self.cfg.data.model / ds_name
```
lauritowal, line 50: I think we would need to do the same for models too.
However, I feel like, what we want instead maybe, is to be able to set a custom path for the whole root ... or at least what follows after elk_reporter_dirt()

lauritowal (changes requested):

norabelrose:
hmm it's not really clear to me that we need this in addition to the custom `out_dir`

PR #188, opened 2023-04-13
fix multi eval behavior

ChristyKoh:
realized evals were saving identical values -- needed to update Extract to only use a single dataset when looping through.


lauritowal:
@ChristyKoh is this done or are you planning to change something, since it's still a draft

lauritowal (approved):

lauritowal (changes requested):
Please resolve merge conflicts

PR #196, opened 2023-04-17
Fix acc for `supervised` (in the same way as #195)

AlexTMallen:


norabelrose:
This should be fixed by #197

PR #197, opened 2023-04-17
Cluster bootstrap for metrics; refactor metric computations into evaluate_preds

norabelrose:
I realized that our code for computing ROC AUROC, accuracy, and calibrated accuracy were sort of all over the place and there was a decent amount of code duplication. This PR refactors all of that into a single function `evaluate_preds` which is used for both reporters and logistic regression classifiers, in `elicit` as well as `eval`.

Other changes:
1. Confidence intervals now use the _cluster bootstrap_, resampling entire groups of prompt templates at a time, to take account of the fact that different variants of the same data point are not IID. This leads to significantly larger CIs than those reported in `main`.
2. Partially in order to account for (1), I've increased the default for `max_examples` from `[750, 250]` to `[1000, 1000]`. There's just too much noise in the data with only 250 clusters.
3. Confidence intervals are now included for accuracy and calibrated accuracy


AlexTMallen (approved):
LGTM

PR #202, opened 2023-04-19
Support non-fast tokenizers like LLaMA

norabelrose:
Unfortunately our current method of getting the final layer LM predictions relies on the `return_offsets_mapping` flag on the tokenizer, which isn't supported for "non-fast" tokenizers (ones implemented in Python and not Rust). The LLaMA tokenizer is currently implemented in Python, as well as the UnifiedQA v2 tokenizer.

This PR changes the way that we concatenate questions and answers together, concatenating them _after_ tokenization rather than before, so that we can support the slow tokenizers.


elk/extraction/extraction.py
```diff
@@ -163,54 +174,55 @@ def extract_hiddens(
...
-                if model.config.is_encoder_decoder:
-                    # TODO: Maybe make this more generic for complex templates?
-                    text = text[:answer_start].rstrip()
-                    target = choice["answer"]
-                else:
-                    target = None
+                target = choice["answer"] if is_enc_dec else None
 
                 # Record the EXACT string we fed to the model
                 variant_inputs.append(text)
```
AlexTMallen, line 183: This no longer records the exact text input that was fed into the model

elk/extraction/extraction.py
```diff
@@ -163,54 +174,55 @@ def extract_hiddens(
...
+                    answer = assert_type(Tensor, encoding.labels)
+                else:
+                    encoding2 = tokenizer(
+                        choice["answer"],
+                        add_special_tokens=False,
+                        return_tensors="pt",
+                    ).to(device)
+                    answer = assert_type(Tensor, encoding2.input_ids)
+
+                    input_ids = torch.cat([input_ids, answer], dim=-1)
```
AlexTMallen, line 203: I don't believe `question` ends with whitespace nor `answer` begins with whitespace, so isn't the input text slightly malformed?

AlexTMallen (changes requested):
I seems like the input to the language model isn't separated properly, but I can't easily check because the text input to the language model isn't stored (just the question part of it).

elk/extraction/extraction.py
norabelrose: No, because the tokenizer doesn't explicitly represent the whitespace anyway. For example, "hello world" is just represented as a token for "hello" followed by a token for "world" in basically every tokenizer I can think of. FWIW `lm-eval-harness` does make the assumption that this is valid to do

AlexTMallen (approved):

Issue #186, opened 2023-04-13
Smoke tests for elk eval command


norabelrose:
Currently we have smoke tests for elk elicit but not elk eval

thejaminator:
I am the test master!

ChristyKoh:
I'd like to learn to write tests too!

thejaminator:
> I'd like to learn to write tests too!

Hey @ChristyKoh! 

n all seriousness, you can go ahead and try writing them, there's an existing smoke test for elicit that you can follow. The smoke test currently simply checks for files being created - it's quite simple. 
I'll be busy applying for SERIMATS these two days so ill probably only get started later anyways. So you can ping me and i can help to review your PR once it's ready

derpyplops:
Fixed and closed by #204

PR #204, opened 2023-04-20
Eval tests

ChristyKoh:
Addresses #186

### DIFF
--- a/elk/extraction/dataset_name.py
+++ b/elk/extraction/dataset_name.py
@@ -1,7 +1,6 @@
-from dataclasses import dataclass
 from typing import NamedTuple
 
-from datasets import DatasetDict, DatasetInfo, get_dataset_config_info
+from datasets import DatasetDict
 
 
 def extract_dataset_name_and_config(dataset_config_str: str) -> tuple[str, str]:


tests/test_smoke_eval.py
```diff
@@ -0,0 +1,120 @@
+from pathlib import Path
+
+from elk import Extract
+from elk.evaluation import Eval
+from elk.extraction import PromptConfig
+from elk.training import CcsReporterConfig, EigenReporterConfig
+from elk.training.train import Elicit
+from tests.utils import *
+
```
thejaminator, line 9: seems you have some uncommited file utils

ChristyKoh: whoops, I just pushed!

ChristyKoh:
The tests are structured as follows:
1. `setup_elicit` creates and runs elk elicit with desired model/dataset
2. `eval_run` creates and runs eval, can specify transfer datasets and ccs/vinc
3. `eval_assert_files_created` checks:
  - vanilla eval results in modified eval.csv
  - transfer eval creates new eval subdirectory structure and files
  
Does this sound reasonable? It's currently failing to assert directory creation for transfer evals.

thejaminator:
> The tests are structured as follows:
> 
> 1. `setup_elicit` creates and runs elk elicit with desired model/dataset
> 2. `eval_run` creates and runs eval, can specify transfer datasets and ccs/vinc
> 3. `eval_assert_files_created` checks:
> 
> * vanilla eval results in modified eval.csv
> * transfer eval creates new eval subdirectory structure and files
> 
> Does this sound reasonable? It's currently failing to assert directory creation for transfer evals.

taking a look at it!

elk/files.py
```diff
@@ -38,3 +38,8 @@ def memorably_named_dir(parent: Path):
     out_dir = parent / sub_dir
     out_dir.mkdir(parents=True, exist_ok=True)
     return out_dir
+
+
+def transfer_eval_directory(source: str) -> Path:
+    """Return the directory where transfer evals are stored."""
+    return elk_reporter_dir() / source / "transfer_eval"
```
thejaminator: because tests need to know where it is
(and also we previously mispelled it as `transfer_evals` (with an s) rather than `transfer_eval`, haha

tests/test_smoke_eval.py
```diff
@@ -1,22 +1,16 @@
...
 from elk.training.train import Elicit
 
-ELICIT_EXPECTED_FILES = [
-    "cfg.yaml",
-    "fingerprints.yaml",
-    "lr_models",
-    "reporters",
-    "eval.csv",
-]
-
```
thejaminator: i removed the elicit assertions cos thats handled by the elicit smoke tests

tests/test_smoke_eval.py
```diff
@@ -56,7 +49,7 @@ def check_contains_files(dir: Path, expected_files: list[str]):
         assert file in created_file_names
 
 
-def eval_run(elicit: Elicit, tfr_datasets: list[str] = None) -> int:
+def eval_run(elicit: Elicit, transfer_datasets: Sequence[str] = []) -> int:
```
thejaminator, line 52: btw its a sequence instead of a list cos list is mutable, so pyright will complain about mutable defaults. but with a protocol of sequence, pyright will complain if you try and mutate it, so that mutable default warning goes awa

tests/test_smoke_eval.py
```diff
@@ -66,51 +59,42 @@ def eval_run(elicit: Elicit, tfr_datasets: list[str] = None) -> int:
...
+        extract.prompts.datasets = transfer_datasets
 
     eval = Eval(data=extract, source=tmp_path)
     eval.execute()
     return start_time_sec
 
 
-def eval_assert_files_created(elicit: Elicit, start_time_sec=0):
+def eval_assert_files_created(elicit: Elicit, transfer_datasets: Sequence[str] = []):
     tmp_path = elicit.out_dir
```
thejaminator, line 72: eval's execute function, on the latest branch, will only run for the specified transfer datasets so theres no self eval now

tests/test_smoke_eval.py
ChristyKoh: I see, thanks for explaining!!

elk/extraction/extraction.py
```diff
@@ -275,14 +279,14 @@ def extract(
...
+) -> DatasetDictWithName:
     """Extract hidden states from a model and return a `DatasetDict` containing them."""
 
     def get_splits() -> SplitDict:
         available_splits = assert_type(SplitDict, info.splits)
         train_name, val_name = select_train_val_splits(available_splits)
 
-        pretty_name = colorize(assert_type(str, info.builder_name), highlight_color)
+        pretty_name = colorize(assert_type(str, ds_name), highlight_color)
         print(
```
thejaminator, line 290: previously `info.builder_name` can sometimes be none, e.g. `christykoh/imdb` has the builder name somehow.. set to None. this caused an exception for such cases.

elk/utils/data_utils.py
```diff
@@ -26,24 +26,6 @@ def get_columns_all_equal(dataset: DatasetDict) -> list[str]:
...
-    config_name, *rest = [ds.config_name for ds in dataset.values()]
-    if not all(name == config_name for name in rest):
-        raise ValueError(
-            f"All splits must have the same config name; got {[config_name, *rest]}"
-        )
-
-    include_config = config_name and has_multiple_configs(builder_name)
-    return builder_name + " " + config_name if include_config else builder_name
-
-
```
thejaminator, line 46: also we don't need this anymore since we simply have the name in DatasetDictWithName. yay!

elk/extraction/extraction.py
thejaminator: theres some things i tried, like setting info.builder_name to the path directly if it doesn't exist.

but theres some trippy shizzles going on (as you may have encountered before) with the builder name.

e.g. `builder.download_and_prepare` would reset the builder_name to the original info, even though we set it to something else in `_GeneratorBuilder`. This was quite annoying and hard to reason with, so i decided to just put the name in `DatasetDictWithName`

elk/evaluation/evaluate.py
```diff
@@ -20,8 +20,12 @@ class Eval(Run):
...
         assert self.source, "Must specify a source experiment."
 
-        transfer_dir = elk_reporter_dir() / self.source / "transfer_eval"
-        self.out_dir = transfer_dir / "+".join(self.data.prompts.datasets)
+        # Set the output directory to the transfer directory if it's not specified
+        self.out_dir = (
+            transfer_eval_directory(self.source)
+            if self.out_dir is None
+            else self.out_dir
+        )
```
thejaminator, line 28: previously was wrongly specified to be `transfer_dir / "+".join(self.data.prompts.datasets)`. 
Also i only set it if it wasn't already specified. Seems like sweep wants to specify their own directory so I didn't want to always override it.

PR #210, opened 2023-04-22
Clean up train and eval commands; do transfer eval in sweep

norabelrose:
Merged `Train` and `Evaluate` classes into their corresponding config classes; removed duplicate code; made the sweep command also run transfer eval on all pairs


lauritowal (approved):
looks good to me. Thanks for getting rid of the unnecessary classes!

PR #211, opened 2023-04-22
Use separate CSVs for LR, LM, & reporter eval; support INLP

norabelrose:
I wanted to see how many orthogonal _supervised_ classifiers can get above random perf, which is a proxy for the "dimensionality of the truth subspace." To do this I wrote a quick and dirty implementation of [Iterative Nullspace Projection (INLP)](https://arxiv.org/abs/2004.07667). While INLP was originally proposed for erasing concepts, R-LACE is much better for that purpose. But INLP _does_ make sense if you just want a bunch of orthogonal classifiers!

Depends on #210

### DIFF
--- a/elk/training/sweep.py
+++ b/elk/training/sweep.py
@@ -89,11 +89,12 @@ class Sweep:
                     if eval_dataset in train_datasets:
                         continue
 
+                    data = deepcopy(run.data)
+                    data.model = model_str
+                    data.prompts.datasets = [eval_dataset]
+
                     eval = Eval(
-                        data=Extract(
-                            model=model_str,
-                            prompts=PromptConfig(datasets=[eval_dataset]),
-                        ),
+                        data=data,
                         source=str(run.out_dir),
                         out_dir=out_dir,
                     )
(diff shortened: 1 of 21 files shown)


elk/training/sweep.py
```diff
@@ -57,10 +72,29 @@ def execute(self):
...
+                # Now evaluate the reporter on the other datasets
+                for eval_dataset in eval_datasets:
+                    # We already evaluated on this one during training
+                    if eval_dataset in train_datasets:
+                        continue
+
+                    eval = Eval(
+                        data=Extract(
+                            model=model_str,
+                            prompts=PromptConfig(datasets=[eval_dataset]),
```
AlexTMallen, line 95: I think the Extract config here should be initialized with a deepcopy of the base Extract config, because currently `max_examples` and other extraction args don't affect evals. Also, this can help the cache performance of extraction.

AlexTMallen (changes requested):
LGTM other than some issues with Eval args in sweep.

elk/training/sweep.py
norabelrose: Fixed

PR #212, opened 2023-04-23
add pdbpp

lauritowal:
"This module is an extension of the [pdb](http://docs.python.org/library/pdb.html) module of the standard library. It is meant to be fully compatible with its predecessor, yet it introduces a number of new features to make your debugging experience as nice as possible."
The "sticky" feature is quite useful for example
https://pypi.org/project/pdbpp/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,6 +43,7 @@ dev = [
     "pytest",
     "pyright",
     "scikit-learn",
+    "pdbpp"
 ]
 
 [project.scripts]


norabelrose:
Hmm maybe this is good but I've never tried it and strictly speaking it's not a dev _dependency_. The other dev dependencies are needed for the tests to run, or are for pre-commit stuff

lauritowal:
alright, I'll close it then. It's still worth a look :)

PR #215, opened 2023-04-26
Log ensembled metrics

norabelrose:
This PR introduces three "modes" for ensembling the different credences output by a reporter for a single cluster or datapoint.

1. `none` is no ensembling, or what we do right now.
2. `partial` means we ensemble across contrast pairs, similar to what was done in the DLK paper. Specifically, the ensembled credence is `pos - neg` where `pos` is the credence on the positive pseudolabel and `neg` is the credence on the negative pseudolabel. 
3. `full` ensembles both across contrast pairs _and_ across different prompt templates. Specifically, we take the average of the credences computed across all prompt templates and use that as the prediction for the cluster.

These are implemented as options on the `evaluate_preds` function. The `elicit` and `eval` commands log metrics for all 3 values.

It turns out this ensembling can matter quite a bit; `full` is often on the order of 10 percentage points higher AUROC / acc than `none`.

### DIFF
--- a/elk/evaluation/evaluate.py
+++ b/elk/evaluation/evaluate.py
@@ -45,21 +45,32 @@ class Eval(Run):
         for ds_name, (val_h, val_gt, _) in val_output.items():
             meta = {"dataset": ds_name, "layer": layer}
 
-            val_result = evaluate_preds(val_gt, reporter(val_h))
-            row_bufs["eval"].append({**meta, **val_result.to_dict()})
+            val_credences = reporter(val_h)
+            for mode in ("none", "partial", "full"):
+                row_bufs["eval"].append(
+                    {
+                        **meta,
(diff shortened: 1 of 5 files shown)


Benw8888 (approved):
Looks great to me, as long as you are fine with the fact that the cal_acc computation still uses both the positive and negative questions even in the "None" ensembling method.

norabelrose:
> Looks great to me, as long as you are fine with the fact that the cal_acc computation still uses both the positive and negative questions even in the "None" ensembling method.

Ummm yeah I _think_ that is expected behavior because the accuracy is supposed to be like "top 1 accuracy" _for the question_, not accuracy for classifying "Q? A" statements as true or false. So inherently you do need to include the credences for both pseudolabels.

PR #223, opened 2023-04-29
Don't try loading models from the cwd ever

norabelrose:
This fixes a very annoying quirk of HuggingFace `from_pretrained` methods: if you try to load, say, `huggyllama/llama-7b` when there's a folder named `huggyllama/llama-7b` in the current working directory, it'll try to load from that folder instead of the HF Hub (or the relevant cache dir). We basically never want this behavior, and it becomes a problem when you happen to be `cd`ed into the `elk-reporters` directory for example.

This PR fixes this with a little hack: we just move the offending local directory to a temp dir for a short time while we call `Auto<whatever>.from_pretrained`, then move it back. AFAICT there's not a better way to do this.

Also, as a bonus, this PR makes the change that Christy and I talked about like a week ago, where datasets like "super_glue boolq" are now named like super_glue:boolq so that you don't have to wrap it in double quotes on the command line

### DIFF
--- a/elk/extraction/dataset_name.py
+++ b/elk/extraction/dataset_name.py
@@ -5,7 +5,7 @@ from datasets import DatasetDict
 
 def extract_dataset_name_and_config(dataset_config_str: str) -> tuple[str, str]:
     """Extract the dataset name and config name from the dataset prompt."""
-    ds_name, _, config_name = dataset_config_str.partition(" ")
+    ds_name, _, config_name = dataset_config_str.partition(":")
     return ds_name, config_name
 
 
--- a/elk/extraction/extraction.py
+++ b/elk/extraction/extraction.py
@@ -35,9 +35,9 @@ from ..utils import (
     instantiate_model,
     instantiate_tokenizer,
     is_autoregressive,
+    prevent_name_conflicts,
     select_train_val_splits,
     select_usable_devices,
-    temporary_dir_move,
 )
 from .dataset_name import (
     DatasetDictWithName,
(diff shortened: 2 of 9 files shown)


AlexTMallen:
I did `mkdir gpt2` from the elk repo root, and then got the following result:

```
❯ elk elicit gpt2 amazon_polarity --max_examples 10 --debug
Traceback (most recent call last):
  File "/mnt/ssd-2/spar/alexm/miniconda3/envs/elk/bin/elk", line 8, in <module>
    sys.exit(run())
  File "/mnt/ssd-2/spar/alexm/elk/elk/__main__.py", line 27, in run
    run.execute()
  File "/mnt/ssd-2/spar/alexm/elk/elk/__main__.py", line 19, in execute
    return self.command.execute()
  File "/mnt/ssd-2/spar/alexm/elk/elk/run.py", line 50, in execute
    self.datasets = [
  File "/mnt/ssd-2/spar/alexm/elk/elk/run.py", line 51, in <listcomp>
    extract(
  File "/mnt/ssd-2/spar/alexm/elk/elk/extraction/extraction.py", line 310, in extract
    with temporary_dir_move(cfg.model):
  File "/mnt/ssd-2/spar/alexm/miniconda3/envs/elk/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "/mnt/ssd-2/spar/alexm/elk/elk/utils/data_utils.py", line 64, in temporary_dir_move
    path.rename(dest)
  File "/mnt/ssd-2/spar/alexm/miniconda3/envs/elk/lib/python3.10/pathlib.py", line 1234, in rename
    self._accessor.rename(self, target)
OSError: [Errno 18] Invalid cross-device link: 'gpt2' -> '/tmp/tmpwn7ysohc/gpt2'```

AlexTMallen (approved):
`elk elicit gpt2 "glue:qnli" --max_examples 10 --debug` works on my end despite a conflicting directory name and the code LGTM

PR #226, opened 2023-04-30
Lots of simplification and refactoring

norabelrose:
- Merge `PromptConfig` into `Extract`, cleaning up the doc strings
- Makes sure to only extract the validation split when we're doing eval
- Replaced the `label_columns` command line arg with a new entry in the prompt template .yaml which indicates what the label column should be
- Got rid of streaming dataset support because it's ~impossible to verify that the data is shuffled correctly and we never used it
- Removed unused stuff in the promptsource module
- Only print `Using X GPUs` once in the lifetime of a program
- Some other things I'm forgetting probably

### DIFF
--- a/elk/extraction/prompt_loading.py
+++ b/elk/extraction/prompt_loading.py
@@ -1,134 +1,40 @@
 from collections import Counter
-from copy import deepcopy
-from dataclasses import dataclass
-from itertools import zip_longest
 from random import Random
-from typing import Any, Iterator, Literal, Optional
+from typing import Any, Iterator, Literal
 
-from datasets import (
-    Dataset,
-    Features,
-    load_dataset,
-)
-from datasets.distributed import split_dataset_by_node
-from simple_parsing.helpers import Serializable, field
+from datasets import ClassLabel, Dataset, Value, load_dataset
 
 from ..promptsource import DatasetTemplates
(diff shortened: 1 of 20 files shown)


elk/extraction/prompt_loading.py
```diff
@@ -165,29 +64,47 @@ def load_prompts(
...
 
+    label_feature = ds.features[label_column]
+    if isinstance(label_feature, ClassLabel):
+        label_choices = [label_feature.str2int(label) for label in label_feature.names]
+    elif isinstance(label_feature, Value) and label_feature.dtype == "bool":
+        label_choices = [False, True]
+    else:
+        # Which classes are actually present in this split of the dataset?
+        # This is shockingly fast since it uses an optimized Apache Arrow primitive.
+        label_choices = sorted(ds.unique(label_column))
```
AlexTMallen, line 77: Perhaps we should warn that we're assuming all class labels are represented?

AlexTMallen (approved):
LGTM!

elk/extraction/prompt_loading.py
norabelrose: Yes

norabelrose: fixed

PR #231, opened 2023-05-01
Load fp32 models in bfloat16 when possible

norabelrose:
Several models that we'd like to evaluate on, like `bigscience/mt0-xxl` and `allenai/unifiedqa-t5-11b`, have float32 checkpoints but were actually trained in bfloat16 on TPUs. Because they're float32, we get out of memory errors when trying to run inference on them. This PR automatically detects if a checkpoint is (likely) float32 before downloading it, and sets `torch_dtype=torch.bfloat16` iff `torch.cuda.is_bf16_supported()` is True.

Some older models, like `gpt2`, have fp32 checkpoints and were just _trained_ in full precision. But it's nearly impossible for an overflow to occur when running these models in bfloat16, since bf16 has a dynamic range almost equal to that of fp32. There is a bit of precision loss, but empirically neural nets are highly robust to this— as long as there aren't any overflows. So this _should_ be fine. We also print a warning when the downcasting does occur. Maybe we should add a flag to turn off this automatic downcasting, but I haven't included it in this PR for simplicity.

### DIFF
--- a/elk/extraction/extraction.py
+++ b/elk/extraction/extraction.py
@@ -31,7 +31,7 @@ from ..utils import (
     Color,
     assert_type,
     colorize,
-    float32_to_int16,
+    float_to_int16,
     infer_label_column,
     infer_num_classes,
     instantiate_model,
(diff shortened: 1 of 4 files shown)


azhx:
Pulling this and running 
```
elk sweep --models \
    bigscience/mt0-xxl \
    --datasets \
    azhx/counterfact-easy
```
continues to OOM. Did this work under some conditions?

azhx (approved):
I pulled this into my branch and ran elk sweep including bigscience/mt0-xxl and it not longer OOMs

PR #232, opened 2023-05-01
CCS results now as expected

norabelrose:
Before we were (embarrassingly) not applying normalization to hidden states before running evaluation, now we are doing that.

TODO: Do the same thing for VINC eval


AlexTMallen (approved):
🎉

PR #235, opened 2023-05-02
Add hyperparamter sweep to `sweep.py`; Fall back to `eig` when `eigh` fails

AlexTMallen:
This also incorporates my 4x smaller reporters change, and falls back to `torch.linalg.eig` when there's an error with `eigh`.


norabelrose (approved):
LGTM

PR #238, opened 2023-05-03
WIP: Generalizable multi gpu

thejaminator:
Try it out with e.g. 2 gpus.
This 
``
elk elicit huggyllama/llama-65b imdb --num_gpus 2 --gpus_per_model 2 --int8
```

WIP: Stil investigating why my bespoke hardcoded device map seemed to perform better.

### CODE REVIEW

File: elk/utils/multi_gpu.py [hunk 1]
```diff
@@ -0,0 +1,184 @@
...
+        {"": first_device}
+        if device_config.is_single_gpu
+        else create_device_map(
+            model_str=cfg.model,
+            use_8bit=cfg.int8,
+            torch_dtype=torch_dtype,
+            model_devices=device_config,
+            verbose=is_verbose,
+        )
+    )
```
thejaminator, line 59: the device map can be different if e.g., gpus 1-7 are all empty. But gpu 0 happens to have someone else's 5 gb model there. 
Suppose you have 4 workers (2 gpus each)

The first worker will have a different device map from the rest, since it'll take into account the 5gb model memory being used.

The device map can sometimes heavily affect performance. In my experience when u can unlucky like (1.5x slower?). This is because you maybe you'll split the model in a suboptimal spot. Like exactly where the model is going to send much more tensors between the layers.

So ideally all the workers would use the same device map so you can approximate the time take to process the dataset better.

elk/utils/multi_gpu.py
```diff
@@ -0,0 +1,184 @@
...
+    # This is because we instantiated our empty model in (probably) float16
+    # We aren't able to instantiate an empty model in 8bit currently
+    max_memory_used_devices = (
+        {
+            device: max_memory_used_devices[device] * 2
+            for device in max_memory_used_devices
+        }
+        if use_8bit
+        else max_memory_used_devices
+    )
```
thejaminator, line 110: the TLDR is that we need to tell `infer_auto_device_map` that we want to use {"cuda:0": 30gb, "cuda:1": 40gb} and NO OTHER GPUS OR CPU OR DISK

elk/utils/hf_utils.py
```diff
@@ -20,44 +22,59 @@
...
     model_str: str,
-    device: str | torch.device = "cpu",
+    load_in_8bit: bool,
+    is_cpu: bool,
+    torch_dtype: Optional[torch.dtype] = None,
     **kwargs,
 ) -> PreTrainedModel:
     """Instantiate a model string with the appropriate `Auto` class."""
-    device = torch.device(device)
-    kwargs["device_map"] = {"": device}
```
thejaminator, line 30: kwargs["device_map"] = {"": device} will be passed by the caller instead (because for e.g. when instantiating an empty model, we can't pass a device map. otherwise it'll really load the weights and won't be an empty model anymore

elk/utils/hf_utils.py
```diff
@@ -20,44 +22,59 @@
...
-        # converting them doesn't hurt performance.
-        elif fp32_weights and torch.cuda.is_bf16_supported():
-            kwargs["torch_dtype"] = torch.bfloat16
-            print("Weights seem to be fp32, but bf16 is available. Loading in bf16.")
-        else:
-            kwargs["torch_dtype"] = "auto"
+        # If a torch_dtype was not specified, try to infer it.
+        kwargs["torch_dtype"] = torch_dtype or determine_dtypes(
+            model_str=model_str, is_cpu=is_cpu, load_in_8bit=load_in_8bit
+        )
```
thejaminator, line 75: made this change because it previously was setting kwargs even if it was getting passed by the caller of instantiate_model, which confused me

PR #242, opened 2023-05-04
LEACE normalization

norabelrose:
Apparently the original CCS code did this, no one noticed until yesterday, and it makes a significant difference to performance


elk/training/eigen_reporter.py
```diff
@@ -185,15 +171,32 @@ def consistency(self) -> Tensor:
...
-        for i, x in enumerate(hiddens.unbind(2)):
-            self.norm.update(x=x, y=torch.full_like(x[..., 0], i))
+
+        # One-hot indicators for each prompt template
+        prompt_ids = torch.eye(v, device=hiddens.device).expand(n, -1, -1)
+        x_neg, x_pos = hiddens.unbind(2)
+        self.norm.update(
+            x=x_neg,
+            # Independent indicator for each (template, pseudo-label) pair
+            y=torch.cat([torch.zeros_like(prompt_ids), prompt_ids], dim=-1),
```
lauritowal, line 188: @norabelrose Is this what we want to do? This results in y having the shape: `torch.Size([250, 10])` for the test_eigen_reporter() . This leads to make 
```assert y.shape[-1] == c, f"Unexpected number of classes {y.shape[-1]}"```
fail since c is 2 and y.shape[-1] is 10

I wanted to fix it myself, but was not sure about what we actually want there. So maybe at least the comment here is helpful for understanding why the test_eigen_reporter fails

lauritowal (changes requested):

PR #252, opened 2023-06-08
Use concept-erasure implementation of LEACE and SAL

norabelrose:
Now that [concept-erasure](https://github.com/EleutherAI/concept-erasure) is on PyPI, we can outsource our `ConceptEraser` implementation to that repo.

This PR makes LEACE, rather than SAL, the default method for pseudolabel and prompt template normalization. I should probably add a config option to change it though.


elk/training/ccs_reporter.py
```diff
@@ -265,12 +265,12 @@ def fit(self, hiddens: Tensor) -> float:
         self.norm.update(
             x=x_neg,
             # Independent indicator for each (template, pseudo-label) pair
-            y=torch.cat([torch.zeros_like(prompt_ids), prompt_ids], dim=-1),
+            z=torch.cat([torch.zeros_like(prompt_ids), prompt_ids], dim=-1),
```
lauritowal, line 268: fixed; replaced y with z for ccs

lauritowal (approved):

elk/training/eigen_reporter.py
AlexTMallen, line 1: nice :) no need for pseudolabels at inference time

AlexTMallen (approved):
LGTM!

elk/evaluation/evaluate.py
```diff
@@ -40,8 +39,7 @@ def apply_to_layer(
         experiment_dir = elk_reporter_dir() / self.source
 
         reporter_path = experiment_dir / "reporters" / f"layer_{layer}.pt"
-        reporter = Reporter.load(reporter_path, map_location=device)
-        reporter.eval()
```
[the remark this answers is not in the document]
norabelrose: No because `CcsReporter` doesn't actually have any submodules like `nn.BatchNorm` or `nn.Dropout` whose behavior changes due to `eval()`

norabelrose:
> JFI, my probes / reporters now won't load with this PR because I used Reporter.load. https://github.com/EleutherAI/elk/pull/252/files#diff-d08b84a509f043deeb98c9c642f692fffbd1967486738d2ff242b7897eb0b1ae

Sorry about that, we can't really guarantee backward compatibility at this point. You should be able to load the reporters with an older commit and extract the raw weights if necessary.

PR #253, opened 2023-06-08
Add git hash in config

derpyplops:
Useful when you want to compare runs across multiple branches/commits


elk/run.py
```diff
@@ -31,6 +32,18 @@
 )
 
 
+def fetch_git_hash() -> str:
+    try:
+        return (
+            subprocess.check_output(["git", "rev-parse", "HEAD"])
```
norabelrose, line 38: Doesn't this depend on the CWD being the git repo or something? If we want to add this feature we should use `__file__` or something similar to figure out where to run `git`

norabelrose (changes requested):

elk/run.py
derpyplops: Yeah, good catch, thanks. Added a test and made the function work even if run in not-the-git-repo.

elk/run.py
```diff
@@ -31,6 +34,20 @@
...
+        return (
+            subprocess.check_output(
+                ["git", "rev-parse", "HEAD"], cwd=Path(elk.__file__).parent.parent
+            )
+            .decode("ascii")
+            .strip()
+        )
+    except Exception as e:
+        print(f"Could not get git revision hash, stack trace: {e}")
+        return "ERROR_HASH_NOT_FOUND"
```
norabelrose, line 48: Sorry to be so picky about this, but this really shouldn't be any kind of "error" because if the user installs from PyPI then it's expected that there won't be a git hash

derpyplops: Oh ok. I had assumed that wouldn't be an issue as there aren't any users who are doing that now, or am I mistaken?

norabelrose: Clément Dumas tried to do it. And I'd like to get more people using it soon

norabelrose (approved):
After my fix this LGTM. The try-catch didn't actually fix the problem with non-editable installs before but now it does.

PR #258, opened 2023-06-15
Add per prompt norm

lauritowal:


elk/training/burns_norm.py
```diff
@@ -0,0 +1,25 @@
...
+class BurnsNorm(nn.Module):
+    """Burns et al. style normalization  Minimal changes from the original code."""
+
+    def forward(self, x: Tensor) -> Tensor:
+        assert (
+            x.dim() == 3
+        ), f"the input should have a dimension of 3 not dimension {x.dim()}, \
+        current shape of input x: {x.shape}"
+
+        x_mean: Tensor = x - torch.mean(x, dim=0)
```
lauritowal, line 14: Probably not the best way to handle input embeddings right now

elk/training/burns_norm.py
```diff
@@ -0,0 +1,25 @@
...
+        if torch.all(x_mean == 0):
+            # input embeddings entries are identical, which leads to x_mean having only zero entries.
+            return x
```
lauritowal, line 17: Probably not the best way to handle input embeddings right now

derpyplops (approved):
lgtm

elk/training/burns_norm.py
```diff
@@ -0,0 +1,26 @@
...
+    """Burns et al. style normalization  Minimal changes from the original code."""
+
+    def forward(self, x: Tensor) -> Tensor:
+        assert (
+            x.dim() == 3
+        ), f"the input should have a dimension of 3 not dimension {x.dim()}, \
+        current shape of input x: {x.shape}"
+
+        x_mean: Tensor = x - torch.mean(x, dim=0)
+        if torch.all(x_mean == 0):
```
norabelrose, line 15: This is kind of nuts. Let's just use `torch.finfo(x.dtype).eps` in the division to prevent NaNs.

elk/training/ccs_reporter.py
```diff
@@ -262,16 +267,19 @@ def fit(self, hiddens: Tensor) -> float:
...
-            # Independent indicator for each (template, pseudo-label) pair
-            y=torch.cat([torch.zeros_like(prompt_ids), prompt_ids], dim=-1),
-        )
-        self.norm.update(
-            x=x_pos,
-            # Independent indicator for each (template, pseudo-label) pair
-            y=torch.cat([prompt_ids, torch.zeros_like(prompt_ids)], dim=-1),
-        )
+        if self.config.norm == "leace":
+            # type ignore because otherwise throws error, probably bug with pyright
```
norabelrose, line 271: It's not a bug with Pyright, it's due to the fact that the type checker doesn't have a way to reliably determine the class of `self.norm` here. If you do an `isinstance` check on `self.norm` it would work.

norabelrose (changes requested):

elk/training/burns_norm.py
norabelrose: Correct

elk/training/ccs_reporter.py
derpyplops: nice!

elk/training/ccs_reporter.py
```diff
@@ -161,8 +164,7 @@ def reset_parameters(self):
     def forward(self, x: Tensor) -> Tensor:
         """Return the credence assigned to the hidden state `x`."""
         assert self.norm is not None, "Must call fit() before forward()"
-
-        raw_scores = self.probe(self.norm(x)).squeeze(-1)
+        raw_scores = self.probe(x).squeeze(-1)
```
lauritowal, line 167: See https://github.com/EleutherAI/elk/pull/274

AlexTMallen (approved):
LGTM!

elk/training/ccs_reporter.py
```diff
@@ -100,14 +100,20 @@ def __init__(
...
                 1 if cfg.num_layers < 2 else hidden_size,
                 bias=cfg.bias,
                 device=device,
-            ),
-        )
+            )
+        ]
+
+        if self.config.norm == "burns":
+            probe_layers.append(nn.Sigmoid())
```
AlexTMallen, line 113: I still don't understand why you're applying a sigmoid only when the normalization is Burns. All the reporters eventually use a sigmoid. The sigmoids are in `losses.py` and `eval.py`. The `credences` passed around are treated as log-odds and we only apply a sigmoid when trying to get probabilities.

elk/training/ccs_reporter.py
```diff
@@ -166,7 +172,10 @@ def forward(self, x: Tensor) -> Tensor:
         assert self.norm is not None, "Must call fit() before forward()"
 
         raw_scores = self.probe(self.norm(x)).squeeze(-1)
-        return raw_scores.mul(self.scale).add(self.bias).squeeze(-1)
+        if self.config.norm == "leace":
+            return raw_scores.mul(self.scale).add(self.bias).squeeze(-1)
+        else:
```
AlexTMallen, line 177: In case we add other normalizers that use the platt scaling parameters, it might be better to do
```
elif self.config.norm == "burns":
    return raw_scores
else:
    raise ValueError(f"Unknown normalization {self.config.norm}.")

File: elk/training/train.py [hunk 8]
```diff
@@ -83,11 +83,12 @@ def apply_to_layer(
             reporter = CcsReporter(self.net, d, device=device, num_variants=v)
             train_loss = reporter.fit(first_train_h)
 
-            (_, v, k, _) = first_train_h.shape
-            reporter.platt_scale(
-                to_one_hot(repeat(train_gt, "n -> (n v)", v=v), k).flatten(),
-                rearrange(first_train_h, "n v k d -> (n v k) d"),
-            )
+            if self.net.norm == "leace":
```
AlexTMallen, line 86: I would slightly prefer `if self.net.norm != "burns"` here for the same reasons of extensibility.

AlexTMallen (changes requested):
Looks good mostly other than the sigmoid layer

AlexTMallen (approved):
LGTM

elk/training/burns_norm.py
```diff
@@ -0,0 +1,47 @@
...
+            x_normalized: normalized output
+        """
+        num_elements = x.shape[0]
+        x_normalized: Tensor = x - x.mean(dim=0) if num_elements > 1 else x
+
+        if not self.scale:
+            return x_normalized
+        else:
+            std = torch.linalg.norm(x_normalized, dim=0) / torch.sqrt(
+                torch.tensor(x_normalized.shape[0], dtype=torch.float32)
```
norabelrose, line 26: Why do we need to wrap this in a Tensor here? You could do 
```py
torch.linalg.norm(x_normalized, dim=0) / x_normalized.shape[0] ** 0.5
```
This is also going to fail if `x` is not `float32`. Probably not a huge deal but I'd mildly prefer for the code to be agnostic to that

elk/training/burns_norm.py
```diff
@@ -0,0 +1,47 @@
...
+            )
+            assert std.dim() == x.dim() - 1
+
+            # Compute the dimensions over which
+            # we want to compute the mean standard deviation
+            # exclude the first dimension v,
+            # which is the template dimension
+            dims = tuple(range(1, std.dim()))
+
+            avg_norm = std.mean(dim=dims)
```
norabelrose, line 36: This all seems overly complex; couldn't we just use `keepdim=True` here?

norabelrose (commented):
Some minor nitpicks

elk/training/burns_norm.py
lauritowal: I'll try and see if that's truly the same

norabelrose: Cool, if it's not the same no worries, I think there are some edge cases where you do need to do hacks like this, but this case seems simple enough that you should be able to avoid it

elk/training/burns_norm.py
lauritowal: done, thanks

elk/training/burns_norm.py
lauritowal: yeah, you were right, it is the same! By using std.mean(dim=dims, **keepdim=True**)  I could simplify the code and get rid of all the unsqueezes. Thanks! :)

PR #259, opened 2023-06-19
Ensembling layer

lauritowal:


elk/metrics/eval.py
```diff
@@ -41,6 +41,73 @@ def to_dict(self, prefix: str = "") -> dict[str, float]:
         return {**auroc_dict, **cal_acc_dict, **acc_dict, **cal_dict}
 
 
+def calc_auroc(y_logits, y_true, ensembling, num_classes):
```
lauritowal, line 44: add annotation

elk/metrics/eval.py
```diff
@@ -41,59 +41,169 @@ def to_dict(self, prefix: str = "") -> dict[str, float]:
...
-    else:
-        raise ValueError(f"Unknown mode: {ensembling}")
 
-    acc = accuracy_ci(y_true, y_pred)
-    cal_acc = None
-    cal_err = None
-
-    if c == 2:
-        pos_probs = torch.sigmoid(y_logits[..., 1] - y_logits[..., 0])
+def calc_eval_results(y_true, y_logits, ensembling, num_classes) -> EvalResult:
```
lauritowal, line 137: add types

elk/metrics/eval.py
```diff
@@ -41,59 +41,169 @@ def to_dict(self, prefix: str = "") -> dict[str, float]:
...
+        EvalResult: The result of evaluating a classifier containing the accuracy,
+        calibrated accuracies, calibrated errors, and AUROC.
+    """
+    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+    y_logits_means = []
+    y_true = layer_outputs[0][0]["val_gt"].to(device)
+
+    for layer_output in layer_outputs:
+        y_logits = layer_output[0]["val_credences"].to(device)
+        y_logits_means.append(y_logits.mean(dim=1))  # full ensembling
```
lauritowal, line 190: we need to change this depending on the ensembling...

elk/metrics/eval.py
```diff
@@ -41,59 +41,169 @@ def to_dict(self, prefix: str = "") -> dict[str, float]:
...
+    # get logits and ground_truth from middle to last layer
+    middle_index = len(layer_outputs) // 2
+    y_logits_stacked = torch.stack(y_logits_means[middle_index:])
+    # layer ensembling of the stacked logits
+    y_logits_stacked_mean = torch.mean(y_logits_stacked, dim=0)
+
+    return calc_eval_results(
+        y_true=y_true,
+        y_logits=y_logits_stacked_mean,
+        ensembling=ensembling,
```
lauritowal, line 202: rename to prompt_ensembling to avoid confusing with layer_ensembling

elk/run.py
```diff
@@ -190,3 +196,24 @@ def apply_to_layers(
                     df.round(4).to_csv(self.out_dir / f"{name}.csv", index=False)
                 if self.debug:
                     save_debug_log(self.datasets, self.out_dir)
+
+                dfs = []
+
+                for ensembling in Ensembling.all():
```
lauritowal: put in a own function

elk/utils/types.py
```diff
@@ -0,0 +1,11 @@
+from enum import Enum
+
+
+class PromptEnsembling(Enum):
```
lauritowal, line 4: Did we really pick a good name here? Is PromptEnsembling a good name?

derpyplops: It's kind of bad but I don't really know better ones

elk/evaluation/evaluate.py
```diff
@@ -31,7 +32,7 @@ def execute(self, highlight_color: Color = "cyan"):
     @torch.inference_mode()
     def apply_to_layer(
         self, layer: int, devices: list[str], world_size: int
-    ) -> dict[str, pd.DataFrame]:
+    ) -> tuple[dict[str, pd.DataFrame], list[dict]]:
```
derpyplops, line 35: Would suggest to add type aliases, list of dict is not very nice to read

`LayerOutput = list[dict]`

derpyplops: And make a new dataclass `LayerResult` that encapsulates the whole output instead of simply returning a tuple

lauritowal (commented):
Tests run forever on my machine. Need to check what is wrong there.

elk/training/train.py
```diff
@@ -53,7 +54,7 @@ def apply_to_layer(
         layer: int,
         devices: list[str],
         world_size: int,
-    ) -> dict[str, pd.DataFrame]:
+    ) -> tuple[dict[str, pd.DataFrame], list[dict]]:
```
AlexTMallen, line 57: Same comment here regarding return type

elk/run.py
```diff
@@ -178,15 +181,39 @@ def apply_to_layers(
         with ctx.Pool(num_devices) as pool:
             mapper = pool.imap_unordered if num_devices > 1 else map
             df_buffers = defaultdict(list)
-
+            layer_outputs = []
             try:
-                for df_dict in tqdm(mapper(func, layers), total=len(layers)):
-                    for k, v in df_dict.items():
+                for df_dict, layer_output in tqdm(
```
AlexTMallen, line 186: This doesn't write all the appropriate lines for 

❯ elk elicit gpt2 imdb amazon_polarity --max_examples 10 300 --debug --num_gpus 1

There should be evaluation results for both imdb and amazon_polarity in the layer_ensembling_results.csv

AlexTMallen (changes requested):
Mainly just fix the handling of the multidataset case

❯ elk elicit gpt2 imdb amazon_polarity --max_examples 10 300 --debug --num_gpus 1

elk/metrics/eval.py
```diff
@@ -41,59 +41,213 @@ def to_dict(self, prefix: str = "") -> dict[str, float]:
...
+        y_logits = layer_output[0]["val_credences"].to(device)
+        y_logits, y_true, num_classes = prepare(
+            y_logits=y_logits,
+            y_true=layer_outputs[0][0]["val_gt"].to(device),
+            prompt_ensembling=prompt_ensembling,
+        )
+        y_logits_collection.append(y_logits)
+
+    # get logits and ground_truth from middle to last layer
+    middle_index = len(layer_outputs) // 2
```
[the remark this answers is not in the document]
lauritowal: yeah, it makes sense to make it configurable. However, I'm curious, how would you decide which layers to pick?

elk/utils/types.py
derpyplops: Well it's bad because per the comments on `discord:EleutherAI:eliciting-latent-knowledge`
```
full = all the ensembling (over contrast pair and templates)
partial = ensembling only over contrast pair (averaging p+ and 1 - p-)
none = looking only at p(input) according to probe, which is most likely what people would do at inference time to keep it simple
If you want you could also define the fourth ensembling method for ensembling over just the templates but not the contrast pair, but that's not how we defined it (we left that option out)
```
It's not only prompt ensembling that this variable controls. It's also pooling over the contrast pair, maybe warranting a more generic name

elk/metrics/eval.py
```diff
@@ -41,59 +41,213 @@ def to_dict(self, prefix: str = "") -> dict[str, float]:
...
+    """
+    if prompt_ensembling == PromptEnsembling.NONE:
+        auroc = roc_auc_ci(
+            to_one_hot(y_true, num_classes).long().flatten(1), y_logits.flatten(1)
+        )
+    elif prompt_ensembling in (PromptEnsembling.PARTIAL, PromptEnsembling.FULL):
+        # Pool together the negative and positive class logits
+        if num_classes == 2:
+            auroc = roc_auc_ci(y_true, y_logits[..., 1] - y_logits[..., 0])
+        else:
```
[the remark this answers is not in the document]
derpyplops: could you clarify what you mean by this please?

Issue #256, opened 2023-06-15
Cannot run README command `elk elicit microsoft/deberta-v2-xxlarge-mnli imdb`


rusheb:
## Summary
Hi there,

I was attempting to set up the application based on the Quick Start guide in the README. When trying to run the `elk elicit microsoft/deberta-v2-xxlarge-mnli imdb` command, the program terminates with a ValueError.

Can you please help me understand what's going wrong and how I can fix it? 

Thanks!

### Environment
- Macbook Air M1
- OS X Ventura
- Python version 3.11.2

### Commands run
```shell
python -m venv .venv
source .venv/bin/activate
pip install -e ".[dev]"
python --version  # outputs Python 3.11.2
# Note: I get a similar error when running other commands from the README
elk elicit microsoft/deberta-v2-xxlarge-mnli imdb
```

I also tried a couple of suggestions from @KayKozaronek, and got the same error:
- Different model/dataset combo: `elk elicit gpt2 boolq`
- Adding a `--net` arg: `elk elicit microsoft/deberta-v2-xxlarge-mnli imdb --net ccs`

### Error output
```shell
ValueError: mutable default <class 'elk.training.train.Elicit'> for field run_template is not allowed: use default_factory
```
(Full stack trace at the bottom of this post)

## Mitigation
GPT tells me we should fix this by replacing the following property from `sweep.py` with a `default_factory`, but without more context on the project I can't tell whether this is the right solution.
```python
    run_template: Elicit = Elicit(
        data=Extract(
            model="<placeholder>",
            datasets=("<placeholder>",),
        )
    )
```

## Full stack trace
```python
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /Users/rusheb/code/elk/venv/bin/elk:5 in <module>                                                │
│                                                                                                  │
│   2 # -*- coding: utf-8 -*-                                                                      │
│   3 import re                                                                                    │
│   4 import sys                                                                                   │
│ ❱ 5 from elk.__main__ import run                                                                 │
│   6 if __name__ == '__main__':                                                                   │
│   7 │   sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])                         │
│   8 │   sys.exit(run())                                                                          │
│                                                                                                  │
│ /Users/rusheb/code/elk/elk/__main__.py:9 in <module>                                             │
│                                                                                                  │
│    6                                                                                             │
[...]

PR #263, opened 2023-06-27
Fixes 256 - ValueError mutable default

artkpv:
Fixes issue #256 (https://github.com/EleutherAI/elk/issues/256) . Error message:

> ValueError: mutable default <class 'elk.training.train.Elicit'> for field run_template is not allowed: use
default_factory

### DIFF
--- a/elk/training/train.py
+++ b/elk/training/train.py
@@ -36,7 +36,7 @@ class Elicit(Run):
     on the training data. "cv" means to use cross-validation."""
 
     @staticmethod
-    def Default():
+    def default():
         return Elicit(
             data=Extract(
                 model="<placeholder>",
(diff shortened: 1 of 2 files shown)


elk/training/train.py
```diff
@@ -34,6 +35,15 @@ class Elicit(Run):
     cross-validation. Defaults to "single", which means to train a single classifier
     on the training data. "cv" means to use cross-validation."""
 
+    @staticmethod
+    def Default():
```
norabelrose, line 39: nit: method names should always be lower case

norabelrose:
If we need to change this for Python 3.11 I would prefer to just use `functools.partial` or something similar.

elk/training/train.py
artkpv: Fixed. Thanks

norabelrose (approved):

PR #266, opened 2023-07-04
fix argument passthrough for sweep

derpyplops:
fixes not-273

### DIFF
--- a/elk/training/sweep.py
+++ b/elk/training/sweep.py
@@ -162,6 +162,7 @@ class Sweep:
                                     continue
 
                                 assert run.out_dir is not None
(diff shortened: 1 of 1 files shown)


elk/training/train.py
```diff
@@ -48,6 +49,26 @@ def create_models_dir(self, out_dir: Path):
 
         return reporter_dir, lr_dir
 
+    def make_eval(self, model, eval_dataset):
```
lauritowal, line 52: not sure if that's the best place to move the code to, everything else looks good to me.

lauritowal (approved):

PR #270, opened 2023-07-12
Fix seed not exposed

AugustasMacijauskas:
Before the fix, running
```bash
elk elicit gpt2 imdb --data.seed=0 --disable_cache
```
and
```bash
elk elicit gpt2 imdb --data.seed=1 --disable_cache
```
would result in identical results. Moreover, printing the `seed` on the first line of the `load_prompts` function would print `42` in both cases which is the default value. https://github.com/EleutherAI/elk/blob/a88c01a07672321e9cc0ac8d32d702e3437deb5c/elk/extraction/prompt_loading.py#L16

After the fix, running the above commands leads to different results (because different data points are sampled from the original training sets) and the print statement now correctly outputs `0` and `1`, respectively.

### DIFF
--- a/elk/extraction/extraction.py
+++ b/elk/extraction/extraction.py
@@ -192,6 +192,7 @@ def extract_hiddens(
         template_path=cfg.template_path,
         rank=rank,
         world_size=world_size,
+        seed=cfg.seed,
     )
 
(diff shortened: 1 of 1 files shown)


norabelrose (approved):
LGTM

PR #271, opened 2023-07-13
train probe per prompt

derpyplops:
Solves NOT-291


lauritowal (commented):
We'll run the sweeps first and see if it improves anything. If yes we can review this + merge it.

PR #277, opened 2023-07-21
output LM stats for , fix debug logging

### LINKED ISSUE #276: Evaluate LM output performance in `elk eval`
Right now only `elk elicit` emits an `lm_eval.csv` but this would also be useful to have for `eval`

AlexTMallen:
Resolves #276


norabelrose (approved):
LGTM

PR #280, opened 2023-07-25
Fix transfer plot bug

reaganjlee:
Previously, elk would go through ``transfer_dir`` to find eval.csv files instead of how it responsively finds csv files like in ``train_dir``, which this fixes. This also renames the ``get_train_dirs`` to be more clear when using it in both directory situations


derpyplops:
LGTM!

PR #286, opened 2023-08-18
Select metric choice for visualization

reaganjlee:
Previously, the visualization code only allowed for ``auroc_estimate`` to be visualized. This PR adds this functionality, allowing users to specify an optional --metric argument in ``elk plot`` or ``elk sweep``


lauritowal (approved):
looked over the code only, without testing it, but it looks good to me

PR #288, opened 2023-08-23
add platt scaling for burns + fix for leace

lauritowal:
- Fixes platt scaling for Leace in CCS, where even during the training of the reporters the platt scaling parameters are multiplied and added to the raw_scores.

- Adds platt_scaling when using burns normalization

### DIFF
--- a/elk/training/ccs_reporter.py
+++ b/elk/training/ccs_reporter.py
@@ -170,6 +170,7 @@ class CcsReporter(nn.Module, PlattMixin):
         if self._is_training:
             return raw_scores
         else:
+            # only do platt scaling after training the reporters
             platt_scaled_scores = raw_scores.mul(self.scale).add(self.bias).squeeze(-1)
             return platt_scaled_scores
 
--- a/elk/training/ccs_reporter.py
+++ b/elk/training/ccs_reporter.py
@@ -9,6 +9,7 @@ import torch
 import torch.nn as nn
 from concept_erasure import LeaceFitter
 from torch import Tensor
+from typing_extensions import override
 
 from ..parsing import parse_loss
 from ..utils.typing import assert_type
@@ -88,7 +89,6 @@ class CcsReporter(nn.Module, PlattMixin):
         num_variants: int = 1,
     ):
         super().__init__()
-        self._is_training = True
 
         self.config = cfg
         self.in_features = in_features
@@ -130,6 +130,15 @@ class CcsReporter(nn.Module, PlattMixin):
                 )
             )
 
+    @override
+    def parameters(self, recurse=True):
+        parameters = super(CcsReporter, self).parameters(recurse=recurse)
+        for param in parameters:
+            # exclude the platt scaling parameters
+            # kind of a hack for now, we should find probably a cleaner way
+            if param is not self.scale and param is not self.bias:
+                yield param
+
     def reset_parameters(self):
         """Reset the parameters of the probe.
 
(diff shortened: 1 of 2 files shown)


elk/training/train.py
```diff
@@ -83,12 +83,9 @@ def apply_to_layer(
             reporter = CcsReporter(self.net, d, device=device, num_variants=v)
             train_loss = reporter.fit(first_train_h)
 
-            if not self.net.norm == "burns":
-                (_, v, k, _) = first_train_h.shape
-                reporter.platt_scale(
-                    to_one_hot(repeat(train_gt, "n -> (n v)", v=v), k).flatten(),
-                    rearrange(first_train_h, "n v k d -> (n v k) d"),
-                )
+            labels = to_one_hot(train_gt, k)
```
AlexTMallen, line 86: this line doesn't do anything it seems

elk/training/train.py
```diff
@@ -83,12 +83,9 @@ def apply_to_layer(
...
+            labels = repeat(train_gt, "n -> n v k", v=v, k=k)
```
AlexTMallen, line 87: Why do we repeat the labels along the choices dimension? I think this means all dimensions of the one-hot prediction are expected to have the same value.

CCS only supports k=2, so we can just deal with that case after after asserting it true.

I think this line should be `labels = to_one_hot(repeat(train_gt, "n -> (n v)", v=v), 2).flatten()`

elk/training/ccs_reporter.py
```diff
@@ -88,6 +88,8 @@ def __init__(
         num_variants: int = 1,
     ):
         super().__init__()
+        self._is_training = True
```
AlexTMallen, line 91: if self._is_training is false during the call to platt_scale, then it won't work because the platt scaling parameter needs to be used for it to be updated in the backwards pass

AlexTMallen (changes requested):

elk/training/ccs_reporter.py
derpyplops, line 91: 🟡 I think this approach basically works but perhaps the wrapper approach that the EigenFitter/Reporter uses is cleaner

elk/training/train.py
lauritowal: Oh nice catch! Thanks

elk/training/train.py
lauritowal: `labels = to_one_hot(repeat(train_gt, "n -> (n v)", v=v), 2).flatten()`
That wouldn't work, since we dimensions have to be the same for the labels and first_train_h, which is a 3D tensor in the case of CCS

derpyplops (changes requested):

elk/training/ccs_reporter.py
derpyplops: Does it not?
```python
if self._is_training:
    return raw_scores
else:
    platt_scaled_scores = raw_scores.mul(self.scale).add(self.bias).squeeze(-1)
    return platt_scaled_scores
```

Anyway seems simple to test by comparing the results with and without the platt scaling params added

derpyplops: In fact maybe we should just do that and make `CcsReporter.fit()` return a `Reporter`. Maybe need to have add a param that disables the eraser.

lauritowal: Yeah, I removed the _is.training again. 

> In fact maybe we should just do that and make CcsReporter.fit() return a Reporter. Maybe need to have add a param that disables the eraser.

We can take a look at that. Could be maybe a new pull request or the same...

AlexTMallen (approved):
lgtm

elk/training/train.py
```diff
@@ -82,13 +82,9 @@ def apply_to_layer(
...
             train_loss = reporter.fit(first_train_h)
-
-            if not self.net.norm == "burns":
-                (_, v, k, _) = first_train_h.shape
-                reporter.platt_scale(
-                    to_one_hot(repeat(train_gt, "n -> (n v)", v=v), k).flatten(),
-                    rearrange(first_train_h, "n v k d -> (n v k) d"),
-                )
+            labels = to_one_hot(train_gt, k)
+            labels = repeat(labels, "n k -> n v k", v=v)
```
derpyplops, line 86: 🟢 nitpick: bugs like the one made here are the reason I dislike reassigning things to the same variable. Either the name should be changed:
`one_hotted_labels = to_one_hot(train_gt, k)`
or it should just be inlined

PR #289, opened 2023-08-24
Add option to save probabilities

AlexTMallen:
- modifies `prepare_data` to return a `LayerData` object
- adds an option to save the raw probabilities, alongside the texts, to the results dir
- adds `train_lr_eval.csv` to results dir

### DIFF
--- a/elk/evaluation/evaluate.py
+++ b/elk/evaluation/evaluate.py
@@ -7,7 +7,7 @@ import torch
 from simple_parsing.helpers import field
 
 from ..files import elk_reporter_dir
-from ..metrics import evaluate_preds
+from ..metrics import evaluate_preds, get_probs
 from ..run import Run
 from ..utils import Color
 
@@ -31,7 +31,7 @@ class Eval(Run):
     @torch.inference_mode()
     def apply_to_layer(
         self, layer: int, devices: list[str], world_size: int
-    ) -> dict[str, pd.DataFrame]:
+    ) -> tuple[dict[str, pd.DataFrame], dict]:
         """Evaluate a single reporter on a single layer."""
         device = self.get_device(devices, world_size)
         val_output = self.prepare_data(device, layer, "val")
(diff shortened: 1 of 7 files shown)


elk/evaluation/evaluate.py
```diff
@@ -31,7 +31,7 @@ def execute(self, highlight_color: Color = "cyan"):
     @torch.inference_mode()
     def apply_to_layer(
         self, layer: int, devices: list[str], world_size: int
-    ) -> dict[str, pd.DataFrame]:
+    ) -> tuple[dict[str, pd.DataFrame], dict]:
```
lauritowal, line 34: Maybe you could do something similar to here: https://github.com/EleutherAI/elk/pull/259/files#diff-d13b83b80dc8fe2ae73e22669dd7a1a3167a1ae731d341fa96f03a766d877933R37
:green_circle:

Instead of having tuple[dict[str, pd.DataFrame], dict]

But we can also leave it for now, and once we merge our pull-request it will be changed anyway

elk/training/train.py
```diff
@@ -152,37 +175,63 @@ def apply_to_layer(
...
+                        if self.save_probs:
+                            out_probs[ds_name]["lr"][mode][i] = (
+                                get_logprobs(val_lr_credences, mode).detach().cpu()
+                            )
+
+                        row_bufs["train_lr_eval"].append(
+                            {
+                                **meta,
+                                "ensembling": mode,
+                                "inlp_iter": i,
```
lauritowal, line 220: what is inlp here?

elk/training/train.py
```diff
@@ -11,7 +11,7 @@
 from simple_parsing import subgroups
```
lauritowal, line 11: The function apply_to_layer was already a bit long. I think we should refactor it a bit in a second pull-request

elk/training/train.py
```diff
@@ -152,37 +175,63 @@ def apply_to_layer(
...
+                    for i, model in enumerate(lr_models):
+                        model.eval()
+                        val_lr_credences = model(val.hiddens)
+                        train_lr_credences = model(train.hiddens)
+                        if self.save_probs:
+                            out_probs[ds_name]["lr"][mode][i] = (
+                                get_logprobs(val_lr_credences, mode).detach().cpu()
+                            )
+
+                        row_bufs["train_lr_eval"].append(
```
lauritowal, line 216: the names are getting a bit confusing. Maybe we should have a subfolder 
evals/
- trainset_lr.csv
- validationset_lr.csv
- trainset_ccs.csv
- etc.

lauritowal: (Could be also included in a second pull-request)

lauritowal (approved):
I think the function apply_to_layer is getting a bit long and maybe a bit confusing, so we might want to create a refactor pull-request in a second step. Everything seems to work fine, though.

elk/training/train.py
AlexTMallen: Iterated nullspace projection iteration for the logistic regression model.

norabelrose:
included in #292

PR #290, opened 2023-08-31
"None" ensembling for classfication accuracy

derpyplops:
Closes NOT-372

ATM, accuracy for "none" ensembling == "partial" ensembling.

This PR implements a reasonable interpretation of what "No ensembling" would look like for classification accuracy: i.e. for accuracy and calibrated accuracy, use the positive hiddens for inference. I also added logging for `cal_thresh`.

### DIFF
--- a/elk/metrics/eval.py
+++ b/elk/metrics/eval.py
@@ -23,6 +23,8 @@ class EvalResult:
     roc_auc: RocAucResult
     """Area under the ROC curve. For multi-class classification, each class is treated
     as a one-vs-rest binary classification problem."""
+    cal_thresh: float | None
+    """The threshold used to compute the calibrated accuracy."""
 
     def to_dict(self, prefix: str = "") -> dict[str, float]:
         """Convert the result to a dictionary."""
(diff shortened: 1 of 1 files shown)


elk/metrics/eval.py
```diff
@@ -23,6 +23,8 @@ class EvalResult:
     roc_auc: RocAucResult
     """Area under the ROC curve. For multi-class classification, each class is treated
     as a one-vs-rest binary classification problem."""
+    cal_thresh: float | None
```
lauritowal, line 26: Any reason why we are adding cal_trhesh to the eval results :green_circle: ?

lauritowal (approved):

elk/metrics/eval.py
derpyplops: i just thought it would be interesting to see 🤷‍♂️ and seemed easy

AlexTMallen (approved):
I am a bit confused why we're only evaluating the positive examples in our "none" ensembling case--it seems slightly arbitrary, but fine. I will also note that this function has always and continues to handle LM logits incorrectly. LM logits are *log probs*, so it does not make sense to apply a sigmoid to them and gives us slightly different ensembled results. Meanwhile, reporter logits are *log odds*, and applying a sigmoid to log odds gives probabilities.

elk/metrics/eval.py
```diff
@@ -64,7 +72,14 @@ def evaluate_preds(
     else:
         y_true = repeat(y_true, "n -> n v", v=v)
 
-    y_pred = y_logits.argmax(dim=-1)
+    THRESHOLD = 0.5
+    if ensembling == "none":
+        y_pred = y_logits[..., 1].gt(THRESHOLD).to(torch.int)
```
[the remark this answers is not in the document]
derpyplops: Yes by "ATM" I meant "behaviour as described in the main branch.

PR #291, opened 2023-08-31
Add log softmax to model_logits

norabelrose:
Ensures that the language model probabilities are normalized across pseudo-labels


AlexTMallen (approved):
tested and lgtm

PR #292, opened 2023-09-21
MAJOR UPDATE: Remove CCS, VINC

AlexTMallen:
This PR does several major things to the repository:

- It removes support for CCS and VINC, and removes the use of contrast pairs
- The `InferenceServer` class is used to do inference, enabling FSDP (`--fsdp`) inference with much larger models
  - It also only requires loading the model once, and doesn't load the model if the dataset is cached
- Only support LR probes and LM predictions
- LM predictions are taken in response to `statement + suffix` where `statement` is the statement we're extracting hidden states from and `suffix` is a piece of text like "\n\nIs the above statement true or false?" that's optionally specified in the yaml template file
- logprobs are optionally stored with `--save_logprobs`
- support for non-balanced datasets with `--balance False`
- Added testing for:
  - The pipeline that turns the raw HF dataset into the input_ids dataset passed to `InferenceServer`
  - `InferenceServer`
- default template called `"_default"` which takes the `"statement"` column itself


elk/promptsource/templates/_default/templates.yaml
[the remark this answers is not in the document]
AlexTMallen: Print an example to stdout

elk/training/classifier.py
```diff
@@ -185,7 +190,12 @@ def fit_cv(
 
     @classmethod
     def inlp(
```
norabelrose, line 192: I think we should probably just get rid of this. If we really want to explore the half-space of admissible predictors we should probably try the Conical Knowledge Decomposition thing.

elk/training/sweep.py
```diff
@@ -127,52 +110,42 @@ def execute(self):
...
+                )
+                run = replace(self.run_template, data=data, out_dir=out_dir)
+
+                # Add hyperparameter values to output directory if needed
+                assert run.out_dir is not None
+
+                try:
+                    run.execute()
+                except torch.linalg.LinAlgError as e:
+                    print(colorize(f"LinAlgError: {e}", "red"))
```
norabelrose, line 126: Wait we can remove this now bc we're not doing eigendecomposition right?

elk/training/train.py
```diff
@@ -7,181 +7,132 @@
...
+    supervised: Literal["single", "inlp", "cv"] = "single"
     """Whether to train a supervised classifier, and if so, whether to use
     cross-validation. Defaults to "single", which means to train a single classifier
     on the training data. "cv" means to use cross-validation."""
 
+    erase_paraphrases: bool = True
+    """Whether to use LEACE to erase the paraphrase dimensions before training the
+    classifier."""
+
+    max_inlp_iter: int | None = None
```
norabelrose, line 32: let's drop this

elk/training/classifier.py
AlexTMallen: I have actually been using this occasionally, basically it just gives me a higher recall way of searching for good probes. Do you still think we should get rid of it?

elk/training/sweep.py
AlexTMallen: Fortunately yes.

norabelrose (approved):
let's merge this bad boy

Issue #295, opened 2024-01-18
Make into a `pytest.fixture` which was a TODO in `test_smoke_eval.py`


RohitRathore1:
https://github.com/RohitRathore1/elk/blob/84e99a36a5050881d85f1510a2486ce46ac1f942/tests/test_smoke_eval.py#L19C1-L20C35

PR #296, opened 2024-01-18
Made into a `pytest.fixture`

RohitRathore1:
It resolves issue #295


RohitRathore1:
Hi, @norabelrose , it would be wonderful if someone from your team have a look on this PR. Thanks!

